PR #1, opened 2020-03-31
update the README and package.json with dependency information

JoshMiers-FS:


package.json
```diff
@@ -11,6 +11,9 @@
     "react-native",
     "fullstory"
   ],
+  "dependencies": {
+    "@fullstory/babel-plugin-react-native": "*"
+  },
   "peerDependencies": {
     "react": "*",
```
patrick-fs, line 18: Would it be a good idea to use better precision on your version? Is it possible that a major version change in react or react-native could be incompatible with the SDK?

package.json
```diff
@@ -11,6 +11,9 @@
     "react-native",
     "fullstory"
   ],
+  "dependencies": {
+    "@fullstory/babel-plugin-react-native": "*"
```
patrick-fs, line 15: See my note about version precision on your peer dependencies - applies here as well. Can a major version change to the babel plugin break the SDK?

README.md
```diff
@@ -18,6 +18,9 @@ npm i @fullstory/react-native --save
 yarn add @fullstory/react-native
 ```
```
patrick-fs, line 20: We've typically added a "This is a WIP" message to the README until the package is deployed on NPM.

package.json
JoshMiers-FS: Yes, it is possible that would happen but for now we are keeping it open during the beta. After, we might tighten it down if/when we come across issues.

package.json
JoshMiers-FS: I don't think this would be an issue, at least in the short term. Once we leave beta and hit `1.0`, I am planning on changing this to be at least the 1.0 version. For now, however, we don't care about versions during the beta.

README.md
JoshMiers-FS: will do.

README.md
```diff
@@ -2,6 +2,7 @@
 
 [![CircleCI](https://circleci.com/gh/fullstorydev/fullstory-react-native.svg?style=svg)](https://circleci.com/gh/fullstorydev/fullstory-react-native)
 
+Note: This is a WIP.
```
patrick-fs, line 5:
```suggestion
Note: This is a Work in Progress and has yet to be published to NPM.
```

PR #11, opened 2020-07-21
iOS: Move swizzles out of FullStory.framework, and implement fsTagName and fsAttribute.

jwise:
Pairs with jwise/unswizzle-react-native branch in FullStory.framework.

Note that modifying an fsAttribute is an untested codepath -- I don't see a test for that in App.js.  I'll work on testing that out, in parallel.

### DIFF
--- a/ios/FullStory.h
+++ b/ios/FullStory.h
@@ -1,5 +1,6 @@
 #import <React/RCTBridgeModule.h>
 #import <FullStory/FS.h>
+#import <FullStory/FSDelegate.h>
 
 @interface FullStory : NSObject <RCTBridgeModule, FSDelegate>
 
--- a/ios/FullStory.m
+++ b/ios/FullStory.m
@@ -1,4 +1,7 @@
 #import "FullStory.h"
+
+#import <objc/runtime.h>
+
 #import <React/RCTView.h>
 #import <React/RCTViewManager.h>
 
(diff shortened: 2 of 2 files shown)


jwise:
Ah, no, the 'click to swap' does the test I'm looking for -- and that codepath is now tested, and works.  So we're good, I'd say.

JoshMiers-FS:
In `App.js`, click the textview that says "Click to swap - false". It will bounce between false and true and on the playback the selectors should change.

JoshMiers-FS:
The component, in playback, should switch between the following for the selector's ending:
`.fs-swap-false.fs-unmask[data-element="TouchableText"][custom_attr2_false="custom_value2_false"][data-source-file="Text.js"][data-component="Text"][custom_attr1_false="custom_value1_false"]`


`.fs-mask.fs-swap-true[data-element="TouchableText"][data-source-file="Text.js"][custom_attr1_true="custom_value1_true"][data-component="Text"][custom_attr2_true="custom_value2_true"]`

JoshMiers-FS:
This ensures that both `fsClass` and `fsAttribute` get their values cleared before adding new ones.

JoshMiers-FS:
btw, i didn't see that you already found the text view to click on before i wrote all of the above. ignore me...

PR #12, opened 2020-07-22
Add `onReady` and move the NativeProps handlers to this plugin

JoshMiers-FS:
In conjunction with the upcoming FS `1.5.5` release, we are moving the NativeProps receivers to this plugin and out from the FS plugin.

Additionally, this PR adds the FS.onReady API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,6 @@ FullStory.getCurrentSessionURL().then(function(result) {
 ### Using the `fsAttribute` property
 Instead of setting attributes via an FS API method, use the `fsAttribute` property that FullStory's babel plugin adds to every React Native `View`.
 
-More information can be found [here](https://help.fullstory.com/hc/en-us/articles/360043356573-FullStory-for-Mobile-Apps-Privacy-Rules#fs-setattribute).
-
 ```JavaScript
 <Text fsAttribute={{custom_attr1: 'custom_value1', custom_attr2: 'custom_value2'}}>Text element with custom attributes</Text>
 ```
@@ -93,8 +91,6 @@ The 6 built-in `fsClass` string values are:
 * "fs-mask-without-consent"
 * "fs-unmask"
 * "fs-unmask-with-consent"
-
-More information can be found in the table at the end of [this section](https://help.fullstory.com/hc/en-us/articles/360043356573-FullStory-for-Mobile-Apps-Privacy-Rules#fs-setattribute).
 
 ```JavaScript
 <Text fsClass="fs-unmask">Text element that is unmasked</Text>
--- a/README.md
+++ b/README.md
@@ -56,10 +56,53 @@ FullStory.event('Subscribed', {
 });
 ```
 
+### Getting a callback once a session is created
+
+```JavaScript
+FullStory.onReady().then(function(result) {
+    const replayStartUrl = result.replayStartUrl;
+    const replayNowUrl = result.replayNowUrl;
+    const sessionId = result.sessionId;
+  });
+```
+
 ### Generating a session replay link
 
 ```JavaScript
 FullStory.getCurrentSessionURL().then(function(result) {
   const startOfPlayback = result;
 });
+```
+
+### Using the `fsAttribute` property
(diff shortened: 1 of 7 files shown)


JoshMiers-FS:
This should be the final FS API additions that are necessary for our initial RN 1.0 support.

android/src/main/java/com/fullstory/reactnative/FullStoryModule.java
```diff
@@ -43,6 +45,42 @@ public static void setUserVars(ReadableMap userVars) {
...
+                    return;
+                }
+
+                WritableMap map = Arguments.createMap();
+
+                // add the replay start url
+                map.putString("replayStartUrl", sessionData.getCurrentSessionURL());
+
+                // add the replay now url
+                map.putString("replayNowUrl", FS.getCurrentSessionURL(true));
```
patrick-fs, line 73: hey @JoshMiers-FS why use two different access patterns here for `getCurrentSessionURL()`? Can `sessionData.getCurrentSessionURL()` be replaced by `FS.getCurrentSessionURL()`?

JoshMiers-FS: It can be replaced and is functionally equivalent. Since the session data already contains the URL, I decided to use that (and it will be ever-so-slightly faster). If you want me to change it, I can do that.

android/src/main/java/com/fullstory/reactnative/FullStoryNativeProps.java
```diff
@@ -0,0 +1,150 @@
...
+    public static void set_dataElement(View view, String value) {
+        setElementIdentity(view, "data-element", value);
+    }
+
+    @NativeProp(name = "dataSourceFile")
+    public static void set_dataSourceFile(View view, String value) {
+        setElementIdentity(view, "data-source-file", value);
+    }
+
+    public static void setProperty(View view, String propName, Object value) {
```
patrick-fs, line 107: Is this an alternative to setting React props on components? How would this be used by a client?

android/src/main/java/com/fullstory/reactnative/FullStoryModule.java
patrick-fs: is there a `getCurrentSessionURL(true)` and `getCurrentSession()` equivalent on `sessionData`?

android/src/main/java/com/fullstory/reactnative/FullStoryNativeProps.java
JoshMiers-FS: I will add a code comment. There is a secondary way of setting a property value generically in the codebase that i haven't actually been able to get RN to use, but I (safely) added support, just in case it is being used in the wild.

android/src/main/java/com/fullstory/reactnative/FullStoryModule.java
JoshMiers-FS: Not currently. I originally was going to add those but decided not to.

JoshMiers-FS:
@patrick-fs - take a look at the README update

README.md
```diff
@@ -56,10 +56,53 @@ FullStory.event('Subscribed', {
...
 ```JavaScript
 FullStory.getCurrentSessionURL().then(function(result) {
   const startOfPlayback = result;
 });
 ```
+
+### Using the `fsAttribute` property
+Instead of setting attributes via an FS API method, use the `fsAttribute` property that FullStory's babel plugin adds to every React Native `View`.
+
+More information can be found [here](https://help.fullstory.com/hc/en-us/articles/360043356573-FullStory-for-Mobile-Apps-Privacy-Rules#fs-setattribute).
```
patrick-fs, line 80: These links might confuse, since they are relevant to iOS and Android native functions rather than React Native. \

patrick-fs: Otherwise Readme LGTM

JoshMiers-FS: I removed the links for now, until we have a RN-specific doc to point to.

PR #30, opened 2021-09-14
Update React Native dependency declarations

JoshMiers-FS:
Using `^` breaks the newest version of `npm`, which will strictly check for peer dependencies. This change addresses that by using `>=`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,11 +12,11 @@
     "fullstory"
   ],
   "dependencies": {
-    "@fullstory/babel-plugin-annotate-react": "^2.0.0",
-    "@fullstory/babel-plugin-react-native": "^1.0.0"
+    "@fullstory/babel-plugin-annotate-react": ">=2.0.0",
+    "@fullstory/babel-plugin-react-native": ">=1.0.0"
   },
   "peerDependencies": {
     "react": "*",
-    "react-native": "^0.61.0"
+    "react-native": ">=0.61.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "fullstory"
   ],
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fullstory/react-native",
-  "version": "1.0.1",
+  "version": "1.0.2",
```
nfriedly, line 3: If you run the command `npm version major|minor|patch` it will increment this for you, commit that change, create a git tag, and other things can hook into it. 

Not a big deal, just an FYI.

package.json
```diff
@@ -12,11 +12,11 @@
     "fullstory"
   ],
   "dependencies": {
-    "@fullstory/babel-plugin-annotate-react": "^2.0.0",
-    "@fullstory/babel-plugin-react-native": "^1.0.0"
+    "@fullstory/babel-plugin-annotate-react": ">=2.0.0",
+    "@fullstory/babel-plugin-react-native": ">=1.0.0"
```
nfriedly, line 16: I wouldn't change these two. They are direct dependencies, so npm won't ever complain about a mismatch, it will just install the right thing.

nfriedly:
The `peerDependencies` change looks good. I would say skip the direct `dependencies` change. Doing the `version` bump here vs in a separate step is up to you.

package.json
JoshMiers-FS: I executed the patch command and that's how the number was updated.

package.json
JoshMiers-FS: Alright, will revert.

JoshMiers-FS: Done.

nfriedly (approved):

PR #31, opened 2021-11-01
add the ability to invoke FS.log via javascript

JoshMiers-FS:
This PR exposes the FS.log infrastructure to React Native Javascript.

The log method takes an int enum and will default to `Info` if the int is invalid.

Usage:
```
import FullStory from '@fullstory/react-native';

// Exercise the logging
FullStory.log(FullStory.LogLevel.Log, 'Log at Log (on iOS, clamps to Debug)');
FullStory.log(FullStory.LogLevel.Debug, 'Log at Debug');
FullStory.log(FullStory.LogLevel.Info, 'Log at Info');
FullStory.log(FullStory.LogLevel.Warn, 'Log at Warn');
FullStory.log(FullStory.LogLevel.Error, 'Log at Error');
FullStory.log(FullStory.LogLevel.Assert, 'Log at Assert (on Android, clamps to Error)');
FullStory.log(-1, 'Log at -1 (defaults to Info)');
FullStory.log(10, 'Log at 10 (defaults to Info)');
```

### DIFF
--- a/ios/FullStory.m
+++ b/ios/FullStory.m
@@ -86,29 +86,27 @@ RCT_EXPORT_METHOD(log:(nonnull NSNumber *)level message:(NSString *)message)
 	// Warn   = 3,
 	// Error  = 4,
 	// Assert = 5, // Clamps to Error on Android
-	NSNumber *levelNumber;
+	FSEventLogLevel levelValue;
 	switch (level.intValue) {
 		case 0:
 		case 1:
-			levelNumber = @(FSLOG_DEBUG);
+			levelValue = FSLOG_DEBUG;
 			break;
 		case 3:
-			levelNumber = @(FSLOG_WARNING);
+			levelValue = FSLOG_WARNING;
 			break;
 		case 4:
-			levelNumber = @(FSLOG_ERROR);
+			levelValue = FSLOG_ERROR;
 			break;
 		case 5:
-			levelNumber = @(FSLOG_ASSERT);
+			levelValue = FSLOG_ASSERT;
 			break;
 		case 2:
 		default:
 			// default to INFO
-			levelNumber = @(FSLOG_INFO);
+			levelValue = FSLOG_INFO;
 		break;
 	}
-
-	FSEventLogLevel levelValue = (FSEventLogLevel)levelNumber.unsignedCharValue;
 
 	dispatch_async(dispatch_get_main_queue(), ^{
 		[FS logWithLevel:levelValue message:message];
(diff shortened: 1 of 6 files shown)


JoshMiers-FS:
@nfriedly - for the js change
@prashanthsadasivan-fullstory - for the Android change
@nvzqz - for the iOS change

prashanthsadasivan-fullstory (approved):

ios/FullStory.m
```diff
@@ -77,6 +77,44 @@ @implementation FullStory {
...
+			levelNumber = @(FSLOG_ASSERT);
+			break;
+		case 2:
+		default:
+			// default to INFO
+			levelNumber = @(FSLOG_INFO);
+		break;
+	}
+
+	FSEventLogLevel levelValue = (FSEventLogLevel)levelNumber.unsignedCharValue;
```
nvzqz, line 111: There is no need to go through `NSNumber` like in my original code. That was because I used a dictionary, which requires boxed values.

```suggestion
	FSEventLogLevel levelValue;
	switch (level.intValue) {
		case 0:
		case 1:
			levelValue = FSLOG_DEBUG;
			break;
		case 3:
			levelValue = FSLOG_WARNING;
			break;
		case 4:
			levelValue = FSLOG_ERROR;
			break;
		case 5:
			levelValue = FSLOG_ASSERT;
			break;
		case 2:
		default:
			// default to INFO
			levelValue = FSLOG_INFO;
		break;
	}
```

ios/FullStory.m
```diff
@@ -77,6 +77,44 @@ @implementation FullStory {
 	});
 }
 
+RCT_EXPORT_METHOD(log:(nonnull NSNumber *)level message:(NSString *)message)
```
nvzqz, line 80: I would still add a null check for `level` that uses `FSLOG_INFO`. Non-nullability in ObjC isn't enforced at runtime and can be simply bypassed. If `level` is null, this would emit `FSLOG_DEBUG` because `intValue` would return 0.

nvzqz (commented):
For iOS this seems good except for 2 things that should be improved/corrected. Thanks for following through on this!

ios/FullStory.m
JoshMiers-FS: I believe I tried with null and the React Native shims will throw an error before it gets to this method, but I will double check.

ios/FullStory.m
JoshMiers-FS: Done

ios/FullStory.m
JoshMiers-FS: I verified that the React Native infrastructure does the null checks for us

nfriedly (approved):
JS LGTM

PR #34, opened 2022-02-28
Adds type declarations

gunnnnii:
This PR adds type declaration to the package.
Most of these types are copied from the browser SDK https://github.com/fullstorydev/fullstory-browser-sdk/blob/main/src/index.d.ts

I'm not 100% sure if all them are exactly correct, I took an educated guess based on the Android Native module definition and [docs](https://help.fullstory.com/hc/en-us/articles/360052419133-Getting-Started-with-FullStory-React-Native-Capture#01ER0FBDD02AS5FXJ0M3076D5R) for `setUserVars` and `getCurrentSession`. Happy to make any changes if these are not completely solid.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,0 +1,21 @@
+interface UserVars {
+    displayName?: string;
+    email?: string;
+    [key: string]: any;
+}
+
+// API functions that are available as soon as the snippet has executed.
+export function anonymize(): void;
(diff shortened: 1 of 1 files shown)


gunnnnii:
closing since this has been added in https://github.com/fullstorydev/fullstory-react-native/pull/35

PR #35, opened 2022-03-10
Add a types declaration for typescript

JoshMiers-FS:
This PR adds better type info for typescript projects. I verified all of the API methods in our sample test app (not linking to private repo here) on both Android and iOS.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -18,7 +18,7 @@ export const LogLevel = {
 };
 
 interface FullStoryInterface {
-  LogLevel: LogLevel;
+  LogLevel: typeof LogLevel;
   anonymize(): void;
   identify(string, Object): void;
   setUserVars(Object): void;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -2,13 +2,23 @@ import {NativeModules} from 'react-native';
 
 declare const {FullStory} = NativeModules;
 
-type OnReadyResponse = {
+export type OnReadyResponse = {
   replayStartUrl: string;
   replayNowUrl: string;
   sessionId: string;
 };
 
+export const LogLevel = {
+  Log: 0, // Clamps to Debug on iOS
+  Debug: 1,
+  Info: 2, // Default
+  Warn: 3,
+  Error: 4,
+  Assert: 5, // Clamps to Error on Android
+};
+
 interface FullStoryInterface {
+  LogLevel: LogLevel;
   anonymize(): void;
   identify(string, Object): void;
   setUserVars(Object): void;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -2,11 +2,17 @@ import {NativeModules} from 'react-native';
 
 declare const {FullStory} = NativeModules;
 
+type OnReadyResponse = {
+  replayStartUrl: string;
+  replayNowUrl: string;
+  sessionId: string;
+};
+
 interface FullStoryInterface {
   anonymize(): void;
   identify(string, Object): void;
   setUserVars(Object): void;
-  onReady(): Promise;
+  onReady(): Promise<OnReadyResponse>;
   getCurrentSession(): Promise<string>;
   getCurrentSessionURL(): Promise<string>;
   consent(boolean): void;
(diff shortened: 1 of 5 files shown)


src/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+  Info: 2, // Default
+  Warn: 3,
+  Error: 4,
+  Assert: 5, // Clamps to Error on Android
+};
+
+interface FullStoryInterface {
+  LogLevel: LogLevel;
+  anonymize(): void;
+  identify(string, Object): void;
```
gunnnnii, line 23: This should probably be using `Record<string, any>`. `Object` is effectively the same as `any`.
F.x. this type checks: `FS.identify('test', 123)`.

src/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+  Log: 0, // Clamps to Debug on iOS
+  Debug: 1,
+  Info: 2, // Default
+  Warn: 3,
+  Error: 4,
+  Assert: 5, // Clamps to Error on Android
+};
+
+interface FullStoryInterface {
+  LogLevel: LogLevel;
```
gunnnnii, line 21: This is a type error. It should be `LogLevel: typeof LogLevel`. This causes the type of `FS.LogLevel[level]` to resolve to any when used.

 [Playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7wDIQOYeAN2ABs4BeOAbwCg44tsAuOABgBo4B6DuAYSIEMAtmFRwYEOABFgAIwCu2OCjgBLAPIBlGlNkKmARlbaAksgBmEJgCZ2XHWf5yiMbQHV+UZEwDMR2gFEoKGgmABY-OABBVFRgWCYAVltuPiERMQlA4KglZCjkABNglQKqAF8AbioqFWQYOIcEYDgAMSciDXEoAE9TeqhG5upaejxCIiZRgmIqkZwx4ismGG6wYAgzOnnpoiqy6oLgBAEoZqQ0eBaNJjaiDq7euob+Jqqqc-Q4IhwyVo0AOim40BPzs-GQ3XeKE+32wVl+VxBuB2ViRnG4yDkghkcSAA)

JoshMiers-FS: Thanks! I had this as a TODO to fix up and will try out your suggestion.

src/index.d.ts
JoshMiers-FS: I will try this out. I was following https://reactnative.dev/docs/native-modules-android#argument-types and then needed to go back and fill out the TS-specific types. I appreciate the comment and will see if it works in our internal tests.

src/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+  identify(string, Object): void;
+  setUserVars(Object): void;
+  onReady(): Promise<OnReadyResponse>;
+  getCurrentSession(): Promise<string>;
+  getCurrentSessionURL(): Promise<string>;
+  consent(boolean): void;
+  event(string, Object);
+  shutdown(): void;
+  restart(): void;
+  log(number, string): void;
```
nfriedly, line 32: Should this be `LogLevel` instead of `number`?

src/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+
+declare const {FullStory} = NativeModules;
+
+export type OnReadyResponse = {
+  replayStartUrl: string;
+  replayNowUrl: string;
+  sessionId: string;
+};
+
+export const LogLevel = {
```
nfriedly, line 11: relatedly, should this be an enum?

src/index.d.ts
nfriedly: Looking at https://reactnative.dev/docs/native-modules-android#argument-types, maybe not...

nfriedly (approved):
I'd be interested to know if an enum worked here, but either way, I think this looks fine.

(I *think* that if an enum would work here, changing it would be a backwards-compatible change.)

JoshMiers-FS:
> I'd be interested to know if an enum worked here, but either way, I think this looks fine.
> 
> (I _think_ that if an enum would work here, changing it would be a backwards-compatible change.)

It doesn't prevent passing in other types, so I am going to leave it as-is for now and look into further ways to change it.

PR #37, opened 2022-04-18
add our base attributes to the global namespace

JoshMiers-FS:
This PR auto-adds our base attributes to the global TS namespace, so our users won't have to do it manually.


JoshMiers-FS:
I verified that this won't negatively impact an app that already has these defined (I did not get any errors). Once this is in and deployed, I will update our RN getting started guide.

nfriedly (approved):
Yes! 💪

PR #40, opened 2022-06-14
Add additional documentation

mikewli:
Add additional documentation to make the FS React Native SDK much clearer to use and implement

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,27 +2,22 @@
 
 [![CircleCI](https://circleci.com/gh/fullstorydev/fullstory-react-native.svg?style=svg)](https://circleci.com/gh/fullstorydev/fullstory-react-native)
 
-FullStory's React Native plugin exposes access to the FullStory Native Mobile SDK from within a React Native app. More information about the FullStory API can be found at https://developer.fullstory.com.
+FullStory's React Native plugin exposes access to the FullStory Native Mobile SDK from within a React Native app. This plug-in is intended to be used in conjunction with [FullStory for Mobile Apps](https://www.fullstory.com/mobile-apps/) for iOS and/or Android.
 
-This plug-in is intended to be used in conjunction with [FullStory for Mobile Apps](https://www.fullstory.com/mobile-apps/). For more information, please see [this](https://help.fullstory.com/hc/en-us/articles/360052419133) getting started guide. Email mobile-support@fullstory.com for additional help.
+## Quick Links
 
-## Install the React Native plugin
-
-#### with npm
-
-```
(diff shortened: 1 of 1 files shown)


JoshMiers-FS:
We initially discussed with @patrick-fs exactly what to put in here and there was a reason it was sparse.

I'd prefer to not manually copy the API docs into yet another place where we need to update information.

mikewli:
@JoshMiers-FS couldn't we have a disclaimer for that we need to purchase FS for FS functionality, since we give examples of how to use the plugin in the `README.md`?

JoshMiers-FS:
For clarity, the README already had a couple API examples in it, which is most likely why @mikewli added the rest of the API methods. This is a good thing, but I also think we should move ALL API (even the ones that were there prior to this PR) to its own doc that we host. I imagine breaking out the API usage from our "getting started with RN capture" guide is also in order.

patrick-fs:
> We initially discussed with @patrick-fs exactly what to put in here and there was a reason it was sparse.
> 
> I'd prefer to not manually copy the API docs into yet another place where we need to update information.

I agree - the links to the getting started guide and developer.fullstory.com provide great guidance. Let's keep reference and "how to" content at those locations and not include them in the project readme.

Having said that, this project readme does document some things that don't exist on on developer.fullstory.com:

> ### Getting a callback once a session is created
> 
> ```JavaScript
> FullStory.onReady().then(function(result) {
>     const replayStartUrl = result.replayStartUrl;
>     const replayNowUrl = result.replayNowUrl;
>     const sessionId = result.sessionId;
>   });
> ```
> 
> ### Generating a session replay link
> 
> ```JavaScript
> FullStory.getCurrentSessionURL().then(function(result) {
>   const startOfPlayback = result;
> });

I think this would be solved by adding a JavaScript (React Native) option in the language switcher for the Client API reference and including these examples there. WDYT?

luser:
Sounds great! We can always regroup and figure out if we want something better once we have a more concrete plan for the revamped developer portal. @mlaux is currently working on adding Kotlin code samples to the docs, the process for adding React Native should be exactly the same.

luser:
> We can always regroup and figure out if we want something better once we have a more concrete plan for the revamped developer portal.

To be more specific: one thing that I'll be looking for in whatever we choose to power our developer portal is the ability to generate documentation from our various API implementations, so in my ideal case we'll be able to document the React Native API using doc comments in the code here, and publishing those docs to the dev portal with everything else.

JoshMiers-FS (approved):
LGTM!

PR #41, opened 2022-07-01
Ryanwang5574/sc 201717/add fullstory support for expo eas

RyanCommits:
Looking for some feedback on this. Will work on updating the README. Things may change form the feedback, so I'm submitting this without the documentation finished.


RyanCommits:
> Maybe omit the server value setting entirely if the host isn't specified (which would be what our customers would want) rather than set it to blank and be in a known-bad state.

I wasn't exactly sure how to do this without leaving a random blank line. For all the configs that are optional, if the user didn't specify a value, they would leave blank lines. Let me know if there's a better solution or if blank lines are fine.

.gitignore
```diff
@@ -4,7 +4,6 @@
 
 # Xcode
 #
-build/
```
RyanCommits, line 7: We have an important `plugin/build` folder so just deleting this for less confusion, since we don't include xcode builds in this project.

JoshMiers-FS:
> > Maybe omit the server value setting entirely if the host isn't specified (which would be what our customers would want) rather than set it to blank and be in a known-bad state.
> 
> I wasn't exactly sure how to do this without leaving a random blank line. For all the configs that are optional, if the user didn't specify a value, they would leave blank lines. Let me know if there's a better solution or if blank lines are fine.

Blank lines are fine.

package.json
```diff
@@ -8,16 +8,37 @@
   "license": "MIT",
   "main": "src/index.js",
   "types": "src/index.d.js",
+  "files": [
```
RyanCommits, line 11: I deleted .npmignore to use allow-listing. Our package downloaded from npm will now only contain the pertinent files and be a bit leaner.

plugin/build/constants.d.ts
```diff
@@ -0,0 +1 @@
+export declare const FULLSTORY_DEFAULT_VERSION = "1.27.1";
```
RyanCommits, line 1: @JoshMiers-FS This may not be the best solution. Should we make `version` a required configuration property? We'll have our users specifically declare the version of FullStory they want and updating this library will not update their FullStory by accident.

plugin/src/__tests__/fixtures/Podfile
```diff
@@ -0,0 +1,50 @@
+require File.join(File.dirname(`node --print "require.resolve('expo/package.json')"`), "scripts/autolinking")
```
RyanCommits, line 1: These fixtures are copy-pasted from expo's blank templates.
https://github.com/expo/expo/tree/master/templates/expo-template-bare-minimum

plugin/src/withFullStoryAndroid.ts
```diff
@@ -0,0 +1,116 @@
...
+    "android.permission.ACCESS_NETWORK_STATE",
+  ]);
+};
+
+export const addFullStoryMavenRepo = (projectBuildGradle: string) => {
+  return mergeContents({
+    tag: `@fullstory/react-native repositories`,
+    src: projectBuildGradle,
+    newSrc: "maven { url 'https://maven.fullstory.com' }",
+    anchor: /repositories/,
```
RyanCommits, line 26: These `anchors` are simply regex matches that inserts `newSrc` where it finds a match. They may not be the most robust matches and can be improved. I plan to stress test these matches by implementing some common expo plugins and see if they collide and break.

plugin/src/withFullStoryIos.ts
```diff
@@ -0,0 +1,98 @@
...
+  withInfoPlist(config, (config) => {
+    config.modResults.FullStory = {
+      OrgId: org,
+      Host: host,
+    };
+    return config;
+  });
+
+const withBuildPhaseDelegate: ConfigPlugin = (config) =>
+  withXcodeProject(config, (config) => {
```
RyanCommits, line 36: This uses prebuilt config-plugins and doesn't dangerously edit content using regex. Testing this function is a little more involved than others since it returns `ConfigPlugin` object rather than a string. Leaving this uncovered by our snapshot testing for now.

plugin/build/constants.d.ts
JoshMiers-FS: Yes, version and org id should both be required for the expo plug-in.

RyanCommits: Updated

RyanCommits:
> > > Maybe omit the server value setting entirely if the host isn't specified (which would be what our customers would want) rather than set it to blank and be in a known-bad state.
> > 
> > 
> > I wasn't exactly sure how to do this without leaving a random blank line. For all the configs that are optional, if the user didn't specify a value, they would leave blank lines. Let me know if there's a better solution or if blank lines are fine.
> 
> Blank lines are fine.

Updated

mikewli:
@RyanCommits We should probably bump up the version number when we're ready to merge this in. What version number it  should be though might need a discussion.

JoshMiers-FS:
> @RyanCommits We should probably bump up the version number when we're ready to merge this in. What version number it should be though might need a discussion.

Yes, we definitely will change the version . The version bump will be independent of this PR.

.gitignore
JoshMiers-FS, line 7: remind me why this was removed

plugin/build/index.d.ts
```diff
@@ -0,0 +1,7 @@
+export declare type FullStoryPluginProps = {
```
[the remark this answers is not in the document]
RyanCommits: updated

.gitignore
```diff
@@ -45,4 +45,5 @@ captures/
 .project
 
 # Allow for types
-!index.d.ts
\ No newline at end of file
+!index.d.ts
```
JoshMiers-FS, line 48: 👍

README.md
```diff
@@ -41,3 +47,38 @@ Here's an example of importing the SDK in a React Native app.
...
+
+Plugins allow for extra customization by passing in an object with properties. If no extra properties are added, defaults will be used. **Certain properties are required.**
+
+| Property        | Platform      | Required                               | Description                                                 |
+| --------------- | ------------- | -------------------------------------- | ----------------------------------------------------------- |
+| version         | Android & iOS | ✅                                     | FullStory for Mobile Apps plugin version                    |
+| org             | Android & iOS | ✅                                     | Your assigned organization ID                               |
+| host            | Android & iOS | Optional. Defaults to: `fullstory.com` | The server url your sessions are sent to                    |
+| enabledVariants | Android       | Optional. Defaults to: `release`       | Specifies which variants to apply FullStory instrumentation |
+| logLevel        | Android       | Optional. Defaults to: `info`          | Captures any log statements at or above the specified level |
```
JoshMiers-FS, line 84: I _think_ we have a log level that we configure in the Info.plist for iOS. I am verifying...

plugin/src/withFullStoryAndroid.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+const withProjectGradleDelegate: ConfigPlugin<FullStoryPluginProps> = (
+  config,
+  { version }
+) => {
+  return withProjectBuildGradle(config, ({ modResults, ...config }) => {
+    if (modResults.language !== "groovy") {
+      throw new Error(
+        "Cannot configure FullStory in the project gradle because the file is not groovy."
```
JoshMiers-FS, line 53: this will become an issue if/when RN ever changes to use gradle script kotlin, but for now it is fine

README.md
JoshMiers-FS: We should call out the possible levels that are available:
https://help.fullstory.com/hc/en-us/articles/360040596093-Getting-Started-with-Android-Recording#01F5E7XFMG19SNYS77NYETKDMQ

However, we will have to tweak things if this also controls the iOS logging (or if we want a separate iOS logging value)

README.md
```diff
@@ -41,3 +47,38 @@ Here's an example of importing the SDK in a React Native app.
...
+```json
+{
+  "expo": {
+    "plugins": [
+      [
+        "@fullstory/react-native",
+        {
+          "version": "1.28.0",
+          "org": "ABC",
+          "enabledVariants": "all"
```
JoshMiers-FS, line 66: I'd remove `enabledVariants`, since it is optional and Android-only

JoshMiers-FS:
I like this and "approve", but am holding off on an official approval, because we need to deploy a new version of the plugin that is independent of this work.

JoshMiers-FS (approved):
LGTM!

Let's get this into master and get some extra testing on it before we release it.

PR #45, opened 2022-09-30
Fixing Typings

devon94:
The  package file was wrong, referencing a js file when it should be ts. The typing were also incorrect in syntax among some other things.

Repro repo is at https://github.com/devon94/fullstoryrepro

No need to for iOS/Android linking as it is purely TypeScript issues. Babel is also not a part of this.

To reproduce: 

1.  Check out repo
2. Yarn install
3. Run `yarn tsc` (compiles TypeScript)
4. Many Failures

image

After my changes:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "author": "FullStory",
   "license": "MIT",
   "main": "src/index.js",
-  "types": "src/index.d.js",
+  "types": "src/index.d.ts",
   "files": [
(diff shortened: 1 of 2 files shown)


RyanCommits:
Thanks for bringing this up. Closed with [comment](https://github.com/fullstorydev/fullstory-react-native/pull/46#issuecomment-1279245016).

PR #46, opened 2022-10-04
fix types

RyanCommits:
Fix TypeScript compile errors when `skipLibCheck = false`. 

Error reproduction from our customer:
 
> (Using the reproduction repo from the customer: https://github.com/devon94/fullstoryrepro)
> 1. Run yarn install
> 2. Run yarn tsc (compiles TypeScript)
> 3. Observe the output errors

To simulate fix, go into `package.json` and replace with: `"@fullstory/react-native": "fullstorydev/fullstory-react-native#ryanwang5574/sc-216416/address-incorrect-typings-syntax-and-incorrect"`

Navigate into `App.tsx` and play with FullStory. You'll see that TS intellisense/autocomplete works correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "author": "FullStory",
   "license": "MIT",
   "main": "src/index.js",
-  "types": "src/index.d.js",
+  "types": "src/index.d.ts",
   "files": [
     "android",
     "ios",
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,35 +1,37 @@
-import {NativeModules} from 'react-native';
-
-declare const {FullStory} = NativeModules;
-
 export type OnReadyResponse = {
   replayStartUrl: string;
   replayNowUrl: string;
   sessionId: string;
 };
 
-export const LogLevel = {
-  Log: 0, // Clamps to Debug on iOS
-  Debug: 1,
-  Info: 2, // Default
-  Warn: 3,
-  Error: 4,
-  Assert: 5, // Clamps to Error on Android
-};
+export enum LogLevel {
+  Log = 0, // Clamps to Debug on iOS
+  Debug = 1,
+  Info = 2, // Default
+  Warn = 3,
+  Error = 4,
+  Assert = 5, // Clamps to Error on Android
+}
 
-interface FullStoryInterface {
+interface UserVars {
+  displayName?: string;
+  email?: string;
+  [key: string]: any;
+}
+
+declare type FullStoryStatic = {
   LogLevel: typeof LogLevel;
   anonymize(): void;
-  identify(string, Object): void;
-  setUserVars(Object): void;
+  identify(uid: string, userVars?: UserVars): void;
+  setUserVars(userVars: UserVars): void;
   onReady(): Promise<OnReadyResponse>;
   getCurrentSession(): Promise<string>;
   getCurrentSessionURL(): Promise<string>;
-  consent(boolean): void;
-  event(string, Object);
+  consent(userConsents: boolean): void;
+  event(eventName: string, eventProperties: Object): void;
   shutdown(): void;
   restart(): void;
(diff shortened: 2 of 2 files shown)


src/index.d.ts
```diff
@@ -1,35 +1,37 @@
...
+export enum LogLevel {
+  Log = 0, // Clamps to Debug on iOS
+  Debug = 1,
+  Info = 2, // Default
+  Warn = 3,
+  Error = 4,
+  Assert = 5, // Clamps to Error on Android
+}
+
+interface UserVars {
```
JoshMiers-FS, line 16: Not sure this is correct having something other than a map sent in. Have you verified that the displayName and email get propagated properly on the FS when using this?

src/index.d.ts
```diff
@@ -1,35 +1,37 @@
...
+  [key: string]: any;
+}
 
-interface FullStoryInterface {
+declare type FullStoryStatic = {
   LogLevel: typeof LogLevel;
   anonymize(): void;
-  identify(string, Object): void;
-  setUserVars(Object): void;
+  identify(uid: string, userVars?: UserVars): void;
```
JoshMiers-FS, line 25: What happens to existing ts code that was passing in an Object?

src/index.d.ts
```diff
@@ -1,35 +1,37 @@
...
+  setUserVars(userVars: UserVars): void;
   onReady(): Promise<OnReadyResponse>;
   getCurrentSession(): Promise<string>;
   getCurrentSessionURL(): Promise<string>;
-  consent(boolean): void;
-  event(string, Object);
+  consent(userConsents: boolean): void;
+  event(eventName: string, eventProperties: Object): void;
```
JoshMiers-FS, line 31: If we are declaring userVars, maybe we can also declare the event properties with a type other than Object as well.

package.json
```diff
@@ -7,7 +7,7 @@
   "author": "FullStory",
   "license": "MIT",
   "main": "src/index.js",
-  "types": "src/index.d.js",
+  "types": "src/index.d.ts",
```
JoshMiers-FS, line 10: Crazy that this made it in. I am wondering how the types file was loaded because I definitely saw changes to the ts intellisense when changing `index.d.ts` locally.

src/index.d.ts
RyanCommits: The `UserVars` interface allows for any generic Object because of this line: `[key: string]: any;`. We added `displayName` and `email` just to document that those two properties are expected properties per our own FS documentation.

src/index.d.ts
RyanCommits: Yes `displayName` and `email` work correctly when using `setUserVars` and `identify`

src/index.d.ts
```diff
@@ -1,35 +1,37 @@
...
   shutdown(): void;
   restart(): void;
-  log(number, string): void;
+  log(logLevel: LogLevel, message: string): void;
```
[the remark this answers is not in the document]
RyanCommits: Nothing changes. Our users can continue using `FullStory.Log(LogLevel.Error, 'message')` or `FullStory.Log(4, 'message')`

devon94:
Just curious, was there something wrong with my PR?

JoshMiers-FS (approved):
LGTM. Once this is in I am going to verify it with our internal sample apps and then create a new release to publish.

RyanCommits:
> Just curious, was there something wrong with my PR?

Nothing wrong with it at all. I chose to declare FullStory as a constant instead of a namespace here, I think that's the major difference.

On the mobile team we don't have an established process (yet) for working with the community. For issue tracking and fixes, we go through our Support team rather than on github. I'm sorry we've lost your commit attribution with this PR, but in order for me to make edits on your work, we need a process for a back and forth on github, and right now we don't have that in place.

PR #48, opened 2022-10-27
CircleCI deploy from tags rather than master

nfriedly:
This updates our CircleCI workflow to run build/test on `v*` branches (e.g. `v1.1.0`) in addition to all branches, and then it moves the hold and deployment jobs to *only* run on `v*` branches.

This is loosely based on https://circleci.com/blog/publishing-npm-packages-using-circleci-2-0/

This should resolve the issue we're currently experiencing where the CI status badge is marked as failing even though we've fixed the issue.

Unfortunately, this stuff is difficult to test until we're ready to do an actual deployment, but I think I got it right.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,13 +41,26 @@ workflows:
   test-deploy:
     jobs:
       - build-react-native
+          filters:
+             # test on branches and version tags (all branches are included by default)
+            tags:
+              only: /^v.*/
       - hold:
           type: approval
           requires:
             - build-react-native
           filters:
+            # hold only on version tags
+            tags:
+              only: /^v.*/
             branches:
-              only: "master"
+              ignore: /.*/
       - deploy-react-native:
           requires:
             - hold
+          filters:
+            # deploy only on version tags
+            tags:
+              only: /^v.*/
+            branches:
+              ignore: /.*/


.circleci/config.yml
```diff
@@ -41,13 +41,26 @@ workflows:
...
           requires:
-            - hold
\ No newline at end of file
+            - hold
+          filters:
+            # deploy only on version tags
+            tags:
+              only: /^v.*/
+            branches:
+              ignore: /.*/
```
nfriedly, line 66: This may be unnecessary since we're depending on `hold` and `hold` already has these filters... but it should be fine like this.

.circleci/config.yml
```diff
@@ -41,13 +41,26 @@ workflows:
   test-deploy:
     jobs:
       - build-react-native
+          filters:
+             # test on branches and version tags (all branches are included by default)
+            tags:
+              only: /^v.*/
```
patrick-fs, line 47: Would it be a good idea to build for all branches, not just ones that have a `v*` tag?

.circleci/config.yml
patrick-fs: We implemented this same configuration for the browser SDK: https://github.com/fullstorydev/fullstory-browser-sdk/blob/49e4b2d7d3f9667befc6bcbdcf329a6aabeee701/.circleci/config.yml#L54

.circleci/config.yml
nfriedly: Do you mean all tags? That probably would be a good idea.

RyanCommits (approved):

nfriedly:
Release process documentation: https://www.notion.so/Release-process-for-React-Native-open-source-packages-292878cf426b404dbfcdd7ec919ad519

JoshMiers-FS (approved):
LGTM

PR #49, opened 2022-11-02
Ryanwang5574/sc 219968/add additional plugin configurations to expo

RyanCommits:
Add additional configurations to expo so that users can further configure Fullstory using the Expo plugin.

Added `logcatLevel`, `recordOnStart`

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,4 +83,3 @@ Plugins allow for extra customization by passing in an object with properties. I
 | enabledVariants | Android       | Optional. Defaults to: `release`       | Specifies which variants to apply FullStory instrumentation                                                |
 | logLevel        | Android       | Optional. Defaults to: `info`          | Captures any log statements at or above the specified level                                                |
 | logcatLevel     | Android       | Optional. Defaults to: `off`           | Captures any Logcat messages at or above the specified level                                               |
-| addDependencies | Android       | Optional. Defaults to: `true`          | Used to prevent FullStory from auto-adding gradle dependencies                                             |
--- a/README.md
+++ b/README.md
@@ -74,10 +74,13 @@ Add the config plugin to the plugins array of your `app.json` or `app.config.jso
 
 Plugins allow for extra customization by passing in an object with properties. If no extra properties are added, defaults will be used. **Certain properties are required.**
 
-| Property        | Platform      | Required                               | Description                                                 |
-| --------------- | ------------- | -------------------------------------- | ----------------------------------------------------------- |
-| version         | Android & iOS | ✅                                     | FullStory for Mobile Apps plugin version                    |
-| org             | Android & iOS | ✅                                     | Your assigned organization ID                               |
-| host            | Android & iOS | Optional. Defaults to: `fullstory.com` | The server url your sessions are sent to                    |
-| enabledVariants | Android       | Optional. Defaults to: `release`       | Specifies which variants to apply FullStory instrumentation |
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -74,10 +74,13 @@ Add the config plugin to the plugins array of your `app.json` or `app.config.jso
...
+| host            | Android & iOS | Optional. Defaults to: `fullstory.com` | The server url your sessions are sent to                                                                   |
+| recordOnStart   | Android & iOS | Optional. Defaults to: `true`          | Setting RecordOnStart to `false` will prevent data capture until you explicitly invoke `FS.restart()` API. |
+| enabledVariants | Android       | Optional. Defaults to: `release`       | Specifies which variants to apply FullStory instrumentation                                                |
+| logLevel        | Android       | Optional. Defaults to: `info`          | Captures any log statements at or above the specified level                                                |
+| logcatLevel     | Android       | Optional. Defaults to: `off`           | Captures any Logcat messages at or above the specified level                                               |
+| addDependencies | Android       | Optional. Defaults to: `true`          | Used to prevent FullStory from auto-adding gradle dependencies                                             |
```
JoshMiers-FS, line 86: I can't see how we would ever support setting `addDependencies` to false, so let's remove this property until/unless we actually need it.

RyanCommits: removed

plugin/src/index.ts
```diff
@@ -9,21 +9,26 @@ import withFullStoryIos from "./withFullStoryIos";
...
-  enabledVariants?: string;
-};
-
-type FullStoryIosProps = {
-  // placeholder
-};
+type LogLevel = "off" | "error" | "warn" | "info" | "debug" | "log";
 
-export type FullStoryPluginProps = {
+type FullStoryCrossPlatformProps = {
```
RyanCommits, line 14: I improved these types by clearly labeling them as Android, iOS, or CrossPlatform, so that `withFullStoryIos` imports only the `iOS` types and `withFullStoryAndroid` imports only the `android` types.

plugin/src/index.ts
```diff
@@ -43,4 +48,4 @@ const withFullStory: ConfigPlugin<FullStoryPluginProps> = (
   ]);
 };
 
-module.exports = createRunOncePlugin(withFullStory, pkg.name, pkg.version);
+export default createRunOncePlugin(withFullStory, pkg.name, pkg.version);
```
RyanCommits, line 51: By changing this to a default export, the resulting file `plugin/build/index.d.ts` is more clear on what type actually is accepted by the config plugin by adding this line: 

`declare const _default: ConfigPlugin<FullStoryPluginProps>;`

I checked with Expo's own plugins: `expo-camera` and `expo-splash-screen`, and they use `export default`. 

Verified that my these changes still work with our Expo test app.

plugin/src/withFullStoryAndroid.ts
```diff
@@ -43,7 +43,7 @@ export const addFullStoryProjectDependency = (
   }).contents;
 };
 
-const withProjectGradleDelegate: ConfigPlugin<FullStoryPluginProps> = (
+const withProjectGradleDelegate: ConfigPlugin<FullStoryAndroidProps> = (
```
RyanCommits, line 46: `FullStoryAndroidProps` is a much accurate type since it would exclude iOS specific props (if there were any)

JoshMiers-FS (approved):
LGTM!

PR #50, opened 2022-11-08
Update package.json

nhnam:
Wrong syntax


RyanCommits:
We support versions greater than `0.61`, syntax is correct.

PR #53, opened 2022-12-08
Update @expo/config-plugins

pettomartino:
Hi team,

I'm updating Expo to use SDK 47, however `expo doctor` complains about outdated version of `@expo/config-plugins`.

Would you accept this update?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.1.1",
       "license": "MIT",
       "dependencies": {
-        "@expo/config-plugins": "^4.1.5",
+        "@expo/config-plugins": "^5.0.4",
         "@fullstory/babel-plugin-annotate-react": "^2.2.0",
         "@fullstory/babel-plugin-react-native": "^1.0.2"
       },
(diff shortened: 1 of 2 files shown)


jlpopken:
Hello all,

We are also using Expo to use SDK 47, and would love for this to be accepted. Thanks!

pettomartino:
This is a little bit more complicated than I thought.
The plugins use `import { mergeContents } from "@expo/config-plugins/build/utils/generateCode";` from version 4 that doesn't seem to be exported in version 5

dereklucas:
Have you had any issues with unmasking while using the resolution override?

jwise:
Hi folks - sorry we missed this for so long :) I confess that we're not quite set up to handle support via Github. Would you mind sending mail to mobile-support@fullstory.com as well, so we can track it through our normal processes and make sure that we don't drop it (and so we can get it linked with your customer accounts, too)?

jwise:
(And, by the way, thanks for this! I'm an iOS native guy so I'm not exactly the right person to review RN stuff but I definitely want to make sure we get the right eyes on this.)

pettomartino:
@dereklucas I haven't noticed any issue.

RyanCommits (changes requested):
Use the recommended dependencies from Expo:
```
{
  "dependencies": {},
  "devDependencies": {
    "expo": "^47.0.0"
  },
  "peerDependencies": {
    "expo": ">=47.0.0"
  },
  "peerDependenciesMeta": {
    "expo": {
      "optional": true
    }
  }
}
```

RyanCommits:
Fixed with https://github.com/fullstorydev/fullstory-react-native/commit/b8af1ce7147c2dc5267272e5da137feeda277587.

PR #55, opened 2023-01-23
send React logs using FS SDK

RyanCommits:
Using this [React Native API](https://github.com/facebook/react-native/blob/3c4850d76b8f67526161211d21a9da66f866e2d6/React/Base/RCTLog.h#L111) to add a custom function that captures every iOS log and sends it to FullStory. 

Is this the best place for the function? Not sure if we can tap into the `AppDelegate` or the lifecycle like `didFinishLaunchingWithOptions` and place it there. Open to suggestions.

There's a question of potential sensitive information capture as well. How should we handle that?

### DIFF
--- a/ios/FullStory.m
+++ b/ios/FullStory.m
@@ -251,7 +251,7 @@ static const char *_rctview_previous_attributes_key = "associated_object_rctview
 
 @implementation LogCapture
 + (void) load {
-	RCTSetLogFunction(^(RCTLogLevel level, RCTLogSource source, NSString *fileName, NSNumber *lineNumber, NSString *message) {
+	RCTAddLogFunction(^(RCTLogLevel level, RCTLogSource source, NSString *fileName, NSNumber *lineNumber, NSString *message) {
         FSEventLogLevel levelValue;
         switch(level) {
             case RCTLogLevelTrace:
--- a/ios/FullStory.m
+++ b/ios/FullStory.m
@@ -245,3 +245,33 @@ static const char *_rctview_previous_attributes_key = "associated_object_rctview
 	} SWIZZLE_END;
 }
 @end
+
+@interface LogCapture : NSObject
+@end
+
+@implementation LogCapture
++ (void) load {
+	RCTSetLogFunction(^(RCTLogLevel level, RCTLogSource source, NSString *fileName, NSNumber *lineNumber, NSString *message) {
+        FSEventLogLevel levelValue;
+        switch(level) {
+            case RCTLogLevelTrace:
+                levelValue = FSLOG_DEBUG;
+                break;
+            case RCTLogLevelInfo:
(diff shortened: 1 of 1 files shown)


ios/FullStory.m
```diff
@@ -245,3 +245,33 @@ + (void) load {
 	} SWIZZLE_END;
 }
 @end
+
+@interface LogCapture : NSObject
+@end
+
+@implementation LogCapture
++ (void) load {
+	RCTAddLogFunction(^(RCTLogLevel level, RCTLogSource source, NSString *fileName, NSNumber *lineNumber, NSString *message) {
```
jwise, line 254: mixed tabs and spaces

ios/FullStory.m
```diff
@@ -245,3 +245,33 @@ + (void) load {
...
+            case RCTLogLevelError:
+            case RCTLogLevelFatal:
+                levelValue = FSLOG_ERROR;
+                break;
+        }
+        
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [FS logWithLevel:levelValue message:message];
+        });
+  });
```
jwise, line 275: inconsistent indendation

ios/FullStory.m
```diff
@@ -245,3 +245,33 @@ + (void) load {
 	} SWIZZLE_END;
 }
 @end
+
+@interface LogCapture : NSObject
+@end
+
+@implementation LogCapture
++ (void) load {
```
jwise, line 253: Having this on a class load is fine conceptually, though I'm curious if there are requirements around when in the app lifecycle `RCTAddLogFunction` is called.  What do other open source APIs do?

jwise:
On web, autocapture of console logs is [behind a switch](https://help.fullstory.com/hc/en-us/articles/360020828533-How-do-I-use-the-FullStory-Console-#01F2HQ0Z0W3HGXC84TXPF0YATZ).  One thing to think about is whether we want to use the same switch to control this... I think there might be a good argument for adding another switch, though, because having this default to on means that we would have a privacy-sensitive behavior change as part of a SDK version bump.

Another challenge of having the switch be serverside-controlled, though, is that we would have to buffer logs until we knew whether we were allowed to pass them along.  (I guess that could be a challenge in late client-side control, too.)

Let's discuss in Slack.

ios/FullStory.m
RyanCommits: Apps typically instruct their users to use `RCTAddLogFunction` themselves and expose a logging function so that the user can call `RCTAddLogFunction` with. Example:

`AppDelegate.m`:
```
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions{
    RCTAddLogFunction(LogFunction);
  ...
}
```

Here the user opts in by adding that function, and not through a static config.

PR #56, opened 2023-01-24
Fix breaking circleCI pipeline and add test job

RyanCommits:
Fix breaking circleCI pipeline and add test job.

1. Needed to upgrade to node16 environment to use `import fs from "fs/promises";`
2. Needed to add TypeScript as dependency to use `ts-jest`.


JoshMiers-FS (approved):
LGTM

PR #58, opened 2023-02-16
WIP: pass props to views that don't

RyanCommits:
```
const CustomComponent = () => {
  return <Text>sample text</Text>
}

class CustomClassComponent extends React.Component {
  render() {
    return <Text>sample text</Text>
  }
}

const FSComponent = withFSAttributes(CustomComponent)
const FSClassComponent = withFSAttributes(CustomClassComponent)

const App = () => {
  return (
    <View>
      <FSComponent fsClass="fs-unmask" />
      <FSClassComponent fsClass="fs-unmask" />
    </View>
  )
}
```

For the above example, `CustomComponent` does not pass props to its children. Using the Higher Order Component `withFSAttributes` will pass `fsClass` and any other props to its children (top level element).

TODO - typescript and tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,8 @@
       "dependencies": {
         "@expo/config-plugins": "^4.1.5",
         "@fullstory/babel-plugin-annotate-react": "^2.2.0",
-        "@fullstory/babel-plugin-react-native": "^1.0.2"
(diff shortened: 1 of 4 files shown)


RyanCommits:
@JoshMiers-FS @prashanthsadasivan-fullstory 
This is a WIP but let me know if this is sufficient to solve our customer's issue.

PR #59, opened 2023-02-19
Add getCurrentSessionURLNow

elliottkember:
When fetching the current session, we need to get the timestamp in the session in order to jump to the correct part of the session.

This PR adds `getCurrentSessionURLNow` to the JS library, calling the native `getCurrentSessionURL` method with a parameter of `true` to get the current timestamp.

### DIFF
--- a/android/src/main/java/com/fullstory/reactnative/FullStoryModule.java
+++ b/android/src/main/java/com/fullstory/reactnative/FullStoryModule.java
@@ -88,10 +88,19 @@ public class FullStoryModule extends ReactContextBaseJavaModule {
         }
     }
 
+    // The original session URL call
     @ReactMethod
     public static void getCurrentSessionURL(Promise promise) {
         if (promise != null) {
             promise.resolve(FS.getCurrentSessionURL());
+        }
+    }
+
+    // The new session URL call - get the current time
+    @ReactMethod
+    public static void getCurrentSessionURLNow(Promise promise) {
+        if (promise != null) {
+            promise.resolve(FS.getCurrentSessionURL(true));
         }
     }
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/fullstory/reactnative/FullStoryModule.java
```diff
@@ -88,13 +88,22 @@ public static void getCurrentSession(Promise promise) {
         }
     }
 
+    // The original session URL call
```
elliottkember, line 91:
```suggestion
```

android/src/main/java/com/fullstory/reactnative/FullStoryModule.java
```diff
@@ -88,13 +88,22 @@ public static void getCurrentSession(Promise promise) {
...
     @ReactMethod
     public static void getCurrentSessionURL(Promise promise) {
         if (promise != null) {
             promise.resolve(FS.getCurrentSessionURL());
         }
     }
 
+    // The new session URL call - get the current time
```
elliottkember, line 99:
```suggestion
```

ios/FullStory.m
```diff
@@ -42,13 +42,22 @@ @implementation FullStory {
 	});
 }
 
+// Original
```
elliottkember, line 45:
```suggestion
```

ios/FullStory.m
```diff
@@ -42,13 +42,22 @@ @implementation FullStory {
...
 RCT_REMAP_METHOD(getCurrentSessionURL, getCurrentSessionURLWithResolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
 {
   dispatch_async(dispatch_get_main_queue(), ^{
 		resolve(FS.currentSessionURL);
 	});
 }
 
+// Added for current session URL at current time
```
elliottkember, line 53:
```suggestion
```

elliottkember (commented):
Self-review

elliottkember:
@JoshMiers-FS Hey, not sure why this check is failing and I don’t have CircleCI access but we still use this and it’s been super helpful.

No pressure to get this merged obviously (we rely on a patch) but would be great to get a review or close this.

PR #60, opened 2023-04-09
fix: expo plugin record on start arg

stezu:
The `recordOnStart` option is true by default and can optionally be disabled. The previous implementation wouldn't set the option in the app's `build.gradle` when provided a `false` value, so this PR fixes that implementation and adds a simple test.


RyanCommits (approved):
LGTM, thanks for your contribution!

RyanCommits:
I merged an update to the dependencies which fixes the build. @stezu if you could rebase off master - the pipeline should pass.

PR #63, opened 2023-06-02
Fix expo plugin record on start arg

RyanCommits:
From #60 ,
The `recordOnStart` option is true by default and can optionally be disabled. The previous implementation wouldn't set the option in the app's `build.gradle` when provided a `false` value, so this PR fixes that implementation and adds a simple test.

Original contributor is inactive. Rebased off master so we can merge his contribution.


JoshMiers-FS (approved):
LGTM

PR #65, opened 2023-06-07
Add support for new architecture

RyanCommits:
Add support for New Architecture through [Turbo Module](https://reactnative.dev/docs/the-new-architecture/backward-compatibility-turbomodules?android-language=java) support. 
Add backwards compatibility with legacy architecture. 

Tested all API methods with Legacy and New Architecture demo apps.


ios/FullStory.mm
```diff
@@ -145,6 +152,24 @@ - (void) fullstoryDidStartSession:(NSString *)sessionUrl {
...
 }
 
+RCT_REMAP_METHOD(onReady, onReadyWithResolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
+{
+	[self onReady:resolve reject:reject];
+}
+
+#ifdef RCT_NEW_ARCH_ENABLED
+- (void)log:(double)logLevel message:(NSString *)message {
+    NSNumber* _Nonnull level = [NSNumber numberWithInt:logLevel];
```
RyanCommits, line 162: Need to manually do a type conversion since `codegen` doesn't translate TypeScript's `number` type correctly.

ios/FullStory.mm
```diff
@@ -56,7 +64,7 @@ @implementation FullStory {
 	});
 }
 
-RCT_EXPORT_METHOD(event:(NSString *)name properties:(NSDictionary *)properties)
+RCT_EXPORT_METHOD(event:(NSString *)name eventProperties:(NSDictionary *)properties)
```
JoshMiers-FS, line 67: Won't changing the name of the argument break existing customer code that uses that argument directly?

JoshMiers-FS: Or is this never called from obj c and only from the code generated from javascript?

android/src/main/java/com/fullstory/reactnative/FullStoryPackage.java
```diff
@@ -1,22 +1,44 @@
...
+import com.facebook.react.module.model.ReactModuleInfoProvider;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
 
-public class FullStoryPackage implements ReactPackage {
+public class FullStoryPackage extends TurboReactPackage {
```
JoshMiers-FS, line 14: Is it always safe to extend TurboReactPackage and not ReactPackage in a legacy app? Why not make the same change you made for FullStoryModule?

ios/FullStory.mm
RyanCommits: This code is never called directly from objc - only from JS, so it won't break on old architecture folks. 

If we ever change these arguments in the future for new architecture, we'll just have to update `src/NativeFullStory.ts` as well.

android/src/main/java/com/fullstory/reactnative/FullStoryPackage.java
RyanCommits: There's two FullStoryModules, and it depends on which architecture the customer is on. 

Old extends [ReactContextBaseJavaModule](https://reactnative.dev/docs/native-modules-android#create-a-custom-native-module-file), new extends [NativeFullStorySpec](https://reactnative.dev/docs/the-new-architecture/pillars-turbomodules#android). I linked examples to the official documentation.

As for `TurboReactPackage`, from the [official docs](https://reactnative.dev/docs/native-modules-android#register-the-module-android-specific):

> You can use [TurboReactPackage](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/TurboReactPackage.java) as an alternative. Instead of createNativeModules, which return a list of instantiated native module objects, TurboReactPackage implements a getModule(String name, ReactApplicationContext rac) method that creates the native module object, when required. TurboReactPackage is a bit more complicated to implement at the moment. In addition to implementing a getModule() method, you have to implement a getReactModuleInfoProvider() method, which returns a list of all the native modules the package can instantiate along with a function that instantiates them, example [here](https://github.com/facebook/react-native/blob/8ac467c51b94c82d81930b4802b2978c85539925/ReactAndroid/src/main/java/com/facebook/react/CoreModulesPackage.java#L86-L165). Again, using TurboReactPackage will allow your application to have a faster startup time, but it is currently a bit cumbersome to write. So proceed with caution if you choose to use TurboReactPackages.

RyanCommits: Here's an example RN library that is backwards compatible with old and new arch.
https://github.com/react-native-webview/react-native-webview/blob/891e595bcf5cafe4a903d5aa643e8361d0c3d467/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewPackage.java#L17

android/src/main/java/com/fullstory/reactnative/FullStoryPackage.java
```diff
@@ -1,22 +1,44 @@
...
+    @Nullable
     @Override
-    public List<NativeModule> createNativeModules(ReactApplicationContext reactContext) {
-        return Arrays.<NativeModule>asList(new FullStoryModule(reactContext));
+    public NativeModule getModule(String name, ReactApplicationContext reactContext) {
+        if (name.equals(FullStoryModuleImpl.NAME)) {
```
JoshMiers-FS, line 18: for null safety, I would swap `FullStoryModuleImpl.NAME` and `name`

android/src/main/java/com/fullstory/reactnative/FullStoryPackage.java
```diff
@@ -1,22 +1,44 @@
...
+        } else {
+            return null;
+        }
     }
 
     @Override
-    public List<ViewManager> createViewManagers(ReactApplicationContext reactContext) {
-        return Collections.emptyList();
+    public ReactModuleInfoProvider getReactModuleInfoProvider() {
+        return () -> {
```
JoshMiers-FS, line 27: This _should_ be OK on all of the AGP plugins we support, since they should all do Java 8 desugaring by default.

android/src/main/java/com/fullstory/reactnative/FullStoryPackage.java
JoshMiers-FS: Thanks. I was concerned about the class not being found but it looks like it was added in 2018 and our minimum RN versions was released in 2019.

RyanCommits:
Updated dependencies to fix vulnerabilities. Retested on iOS & Android on both architectures.

PR #67, opened 2023-06-26
Ryanwang/add pages api

RyanCommits:
Add pages API to RN plugin.
Tested on Android and iOS

https://app.staging.fullstory.com/ui/KWH/session/5735731544391680:(6202668493766656!6238255527428096(3219532248478003708(0:1920:4795:4796:6373:6373:6373:8685:8686:8686)))


android/src/legacy/java/com/FullStoryModule.java
```diff
@@ -86,4 +86,24 @@ public static void log(double level, String message) {
     public static void resetIdleTimer() {
         FullStoryModuleImpl.resetIdleTimer();
     }
+
+    @ReactMethod
+    public static void createPage(String pageName, ReadableMap pageProperties) {
```
JoshMiers-FS, line 91: @ReeceLaF - can you take a look at these changes? I think not propagating the FSPage around might be an issue.

src/index.js
```diff
@@ -30,9 +34,13 @@ const LogLevel = {
...
 };
 
+const createPageWithProperties = (pageName, pageProperties = {}) => createPage(pageName, pageProperties);
+const startPageWithProperties = (pageProperties = {}) => startPage(pageProperties);
+const identifyWithProperties = (uid, userVars = {}) => identify(uid, userVars);
+
 export default {
   anonymize,
-  identify,
+  identify: identifyWithProperties,
```
JoshMiers-FS, line 43: was this missed in the previous PR?

android/src/legacy/java/com/FullStoryModule.java
ReeceLaF: Yeah, ideally this would work the same way as in Android and iOS, and you could retain references to the pages instead of having the library just remember the last one you created. Is it possible to hold a reference to a Java / C object in a wrapper in the RN js? Or if not, would there be a way to notify the Java / C code when the JS object goes out of scope?

src/index.js
RyanCommits: Yes, technically. The typescript type declares `userVars` in identify as optional. Those that will use our Turbomodules will get an error if they don't pass in at least an empty object to `userVars` because the objc code is expecting a `NSDictionary`.

src/FSPage.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.properties = properties;
+    this.nonce = '';
+    this.cleanProperties();
+  }
+
+  private static FS_PAGE_NAME_KEY = 'pageName';
+
+  private static generateUUID() {
+    return 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'.replace(/[x]/g, () => {
+      const char = Math.floor(Math.random() * 16);
```
JoshMiers-FS, line 36: Weird way to generate a UUID. Is this what JS limits us to?

JoshMiers-FS (approved):
Android LGTM

src/FSPage.ts
```diff
@@ -0,0 +1,112 @@
...
+      return char.toString(16);
```
martin-fs, line 37: Note that NSUUIDs are composed of upper case digits, while [Number.toString(radix)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toString) produces [lower case digits](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-number-tostring:~:text=The%20digits%20used%20in%20the%20representation%20of%20a%20number%20using%20radix%20r%20are%20taken%20from%20the%20first%20r%20code%20units%20of%20%220123456789abcdefghijklmnopqrstuvwxyz%22%20in%20order.). There is no documentation regarding the format accepted by [+[NSUUID initWithUUIDString:]](https://developer.apple.com/documentation/foundation/nsuuid/1411662-initwithuuidstring?language=objc) other than an example of "The standard format for UUIDs" which uses uppercase characters. So we should probably call [String.toUpperCase](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase) on the result.

src/FSPage.ts
martin-fs: [Math.random()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) doesn't have to be very random. It could be implemented by a [very low quality PRNG](https://en.wikipedia.org/wiki/RANDU). V8, for example, looked at the statistical quality of their [Math.random implementation in 2015 and improved it](https://v8.dev/blog/math-random). [Hermes' implementation of Math.random()](https://github.com/facebook/hermes/blob/8bc0758c6c95b81074aa4a48c3238be600b2dccf/lib/VM/JSLib/Math.cpp#L209) uses [std::uniform_real_distribution](https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution). The implementation notes indicate that [most implementations for std::uniform_real_distribution will occasionally return the upper end of the range](https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution#:~:text=Most%20existing%20implementations%20have%20a%20bug), so it is possible that the Hermes implementation of Math.random() occasionally returns 1.

Unfortunately, [Hermes does not export Crypto.getRandomValues](https://github.com/facebook/hermes/issues/1003), so I think we have to trust its Math.random() implementation, which, as a recent addition to the C++ standard is probably reasonably high quality, other than the fact that we should explicitly wrap it with Math.min(Math.random(), [0.99999999999999988898](https://www.quora.com/What-is-the-largest-number-less-than-1-that-can-be-represented-using-a-binary64-double-precision-floating-point-register)) (some verification may be needed here)

martin-fs (commented):
I think a couple of changes are needed related to the generation of UUIDs, and we need to do a little verification. Let me know what you think!

src/FSPage.ts
martin-fs: It does appear that LLVM may have fixed [the bug](https://bugs.llvm.org/show_bug.cgi?id=27839) in 2020. (An [older bug ](https://bugs.llvm.org/show_bug.cgi?id=18767) linked from the implementation notes above is still open, but I think this is an oversight.) Not sure Hermes is using the system's libc++, meaning that older devices will have the bug.

Relatedly, [LLVM uses a linear congruential generator](https://en.wikipedia.org/wiki/Linear_congruential_generator#Advantages_and_disadvantages) for std::uniform_real_distribution. LCGs have low statistical quality and it's possible that we're sampling the "less random bits" by doing the Math.floor(random*16). It might be worth generating a bunch of UUIDs and making sure they're all unique.

martin-fs: I think the easiest way to resolve this is to expose a native function that just returns `[[NSUUID alloc] init].UUIDString` and use that.

RyanCommits: Reece and I had a discussion on the tradeoffs between using Math.random() and native UUID generation, for which we updated in the design doc, but here's our conclusion:

Due to the asynchronous nature of how React Native communicates between native code and JavaScript, a challenge to overcome is how to use native code to generate UUIDs. Since a new UUID needs to be generated on page creation, relying on native code to generate the UUID and communicate that to JavaScript will cause page creation to be asynchronous. To prevent asynchronous functionality, we implement a pseudo-random UUID generator on the JavaScript side using Math.random(). The web API ‘crypto’ is not available in React Native. Setting a page’s nonce value to a non cryptographically secure number generator runs the risk of allowing a bad actor with access to an org’s session data to tie users together from a split session using the page’s nonce value. However both the feasibility and likelihood of this risk is low, and functionality of the Pages API does not depend on a cryptographically secure nonce value. Math.random() is therefore sufficient in generating a unique identifier to identify user pages.

src/FSPage.ts
RyanCommits: Added `.toUpperCase()` to UUID generator

src/FSPage.ts
RyanCommits: @martin-fs on the subject of guaranteeing uniqueness of the UUIDs generated - how many are you thinking of generating to establish confidence? In terms of the number of unique pages an org would contain, I don't think our implementation would realistically be in danger of having a duplicate.

RyanCommits: I can this a few times - came back false.

```
function generateUUID() {
  return 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'.replace(/[x]/g, () => {
    const char = Math.floor(Math.random() * 16)
    return char.toString(16).toUpperCase()
  })
}

let oldId = ''
let dupeFound = false

for (let i = 0; i < 1000000; i++) {
  const newId = generateUUID()
  if (oldId === newId) {
    dupeFound = true
  } else {
    oldId = newId
  }
}

console.log('dupe found?', dupeFound)

```

martin-fs: I appreciate you writing a test. Did you check to see whether Hermes may run into the Math.random() == 1 bug?

If it's not feasible to call native code to generate the UUID, I understand. I will want to dig in to understand the Hermes Math.random() implementation on Android and iOS a little more so that we can be reasonably assured of its statistical quality across all devices. Statistical quality is different from cryptographic quality. Check out more about the [advantages and disadvantages](https://en.wikipedia.org/wiki/Linear_congruential_generator#Advantages_and_disadvantages) of LCGs (also linked above)

martin-fs: It turns out [Hermes uses](https://github.com/facebook/hermes/blob/83e798927bbc0390bf11fcac9edd6c446bb4d118/include/hermes/VM/JSLib/RuntimeCommonStorage.h#L41) [std::minstd_rand](https://cplusplus.com/reference/random/minstd_rand/) which is a *very* simple random number generator (an LCG with no additive constant, also known as a Lehmer generator). In addition to the statistical concerns, another issue with this generator is that if seeded with 0, it will only output the value 0. Hermes does not appear to guard against this case.

martin-fs: I did some testing and it looks like both [clang](https://github.com/llvm/llvm-project/blob/358d9dbf4f1b9b3b166139a366d0e8c5162988c0/libcxx/include/__random/linear_congruential_engine.h#L339) and [libstdc++](https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libstdc%2B%2B-v3/include/bits/random.tcc#L121) guard against this fatal initialization condition.

martin-fs: The `Math.floor(Math.random()*16)` operation may use a subset of the generator's bits, which could have a smaller range or cycle than the entire value. In addition, because Math.random() is called 32 times, the generator's state will wrap even more quickly. If a customer calls Math.random() often enough, it is conceivable that the state could wrap. We should not need to tell the customer not to call Math.random() an excess number of times.

I have a proposal below. The code generates non-cryptographic UUIDs using [xorshift128](https://en.wikipedia.org/wiki/Xorshift). This algorithm uses 128 bits of state, so it only wraps after generating 2^128-1 values. This version does four xorshift128 operations in a row, but because 4 is relatively prime to 2^128-1, it will still produce all 2^128-1 values. A single generated UUID, which is also 128 bits, reveals the entire state, but we can be certain that no UUID will be generated twice in one session, and it is extremely unlikely that two sessions will ever generate the same UUID given that the generator is seeded with both Math.random() and Date.now(). (Perhaps we could include other session entropy?)

Note that the minstd generator will also very likely leak its entire state with a single UUID generated, since it only has 32 bits of state, but reverse engineering the state is not *completely* trivial...

Let me know what you think of the proposal below! I'd be happy to compare its performance to the other implementation - it uses 32 bit arithmetic so it should be quite fast.

```javascript
var gen_uuid = function() {

  function hex8(n) {
    return ((n>>>0)+4294967296).toString(16).substring(1).toUpperCase();
  }

  // state (a is always initialized to odd value, so state is never 0)
  var a = (Math.random()*4294967296)|1, b = (Math.random()*4294967296)|0, c = (Date.now() / 4294967296)^(Math.random()*4294967296), d = Date.now()^(Math.random()*4294967296);

  // console.log("Initial state: a: "+hex8(a)+" b: "+hex8(b)+" c: "+hex8(c)+" d: "+hex8(d));
  // four applications of xorshift128
  var f = function() {
    var t = d | 0;
    t ^= t << 11;
    t ^= t >>> 8;
    d = t ^ a ^ (a >>> 19);
    t = c | 0;
    t ^= t << 11;
    t ^= t >>> 8;
    c = t ^ d ^ (d >>> 19);
    t = b | 0;
    t ^= t << 11;
    t ^= t >>> 8;
    b = t ^ c ^ (c >>> 19);
    t = a | 0;
    t ^= t << 11;
    t ^= t >>> 8;
    a = t ^ b ^ (b >>> 19);
    return hex8(a)+"-"+hex8(b).substring(0,4)+"-"+hex8(b).substring(4)+"-"+hex8(c).substring(0,4)+"-"+hex8(c).substring(4)+hex8(d);
  };

  // mix initial state - not strictly necessary
  f();
  f();
  f();
  f();

  return f;
}();

for (var i = 0; i < 100; i++) {
  console.log(gen_uuid());
}
```

RyanCommits: I replaced the code with your implementation, and updated the tests. I also did some manual testing and everything looks fine. Thank you so much for doing the deep dive into this. This code shouldn't get called often so performance is negligible here.

martin-fs (approved):
Thanks for implementing the new UUID generator! (I'm curious what others think about it, as well.)

src/FSPage.ts
ReeceLaF: @martin-fs From my understanding of xorshift128 from the wiki article you linked, I think this:
```javascript
var t = d | 0;
    t ^= t << 11;
    t ^= t >>> 8;
    d = t ^ a ^ (a >>> 19);
```
should actually be this:
```javascript
var t = a | 0;
    t ^= t << 11;
    t ^= t >>> 8;
    d = t ^ d ^ (d >>> 19);
```
I'm not sure if it impacts the soundness of the algo or not, but maybe you would know better if this modification is safe. I think we are also rotating in the opposite direction, which probably doesn't matter but is another difference.

martin-fs: @ReeceLaF We can add a test to verify this x4 implementation against 4 applications of the straightforward implementation. I had gotten the same output in my local testing. @RyanCommits , should I create a branch with a commit you can cherry pick?

ReeceLaF: I'm also wondering why the hex8 function has a 0-bit shift and adds 2**32. Would this not work?
```javascript
function hex8(n) {
    return (n|0).toString(16).toUpperCase();
}
```
If not, a clarifying comment would be great to explain why the other code is needed.

It may also be a good idea to set the version number and variant to conform to the UUID spec.
```javascript
var b2 = b & 0xFFFF0FFF; // clear the version bits
b2 |= 0x00004000; // set the version to 4
var c2 = c & 0x3FFFFFFF; // clear the variant bits
c2 |= 0x80000000; // set variant to IETF
return hex8(a)+"-"+hex8(b2).substring(0,4)+"-"+hex8(b2).substring(4)+"-"+hex8(c2).substring(0,4)+"-"+hex8(c2).substring(4)+hex8(d);
```

martin-fs: Regarding IETF conformance, this is an interesting idea. This was also an issue with the previously proposed JS UUID generation strategy. We aren't generating "real UUIDs" since they are not cryptographic - we are just trying to generate unique values.

We would have to consider whether overwriting those bits would introduce a short cycle in the UUID generation. Currently, we're leveraging the work of Marsaglia, but this will require some novel analysis.

ReeceLaF: I also tested using the original copy-pasted algo in javascript 4 times to create the UUID and saw that it produced the same result. I'm still not sure how it works out to the same value, but it definitely does so it's good with me as is. As for IETF conformance, the UUID doesn't have to be cryptographically secure. Pseudo-random values for v4 are acceptable. It should be fine if the value isn't a "real" UUID though, so I am good to approve now.

ReeceLaF (approved):
Changes LGTM!

JoshMiers-FS:
Great collaboration on the UUID updates @martin-fs @RyanCommits and @ReeceLaF !

PR #68, opened 2023-06-29
Add Android pages implementation

RyanCommits:
Post design doc approval:

- update legacy RN native module method declarations
- replace System.out with android.util.Log
- initialize reflected methods once
- getUUID is async
- fix page properties merge
- add TS FSPage types

RN Fabric [session](https://app.staging.fullstory.com/ui/KWH/session/6630689065336832:5887655561068544)
RN old architecture [session](https://app.staging.fullstory.com/ui/KWH/session/4994683218886656:5840084033536000)

This merges into branch `add-pages-api`. Additional work needed:
- iOS implementation
- add a testing framework to this plugin to test FSPage

### DIFF
--- a/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
+++ b/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
@@ -17,6 +17,7 @@ import java.lang.reflect.Method;
 public class FullStoryModuleImpl {
 
     public static final String NAME = "FullStory";
+    private static final String TAG = "FullStoryModuleImpl";
     public static final boolean reflectionSuccess;
     private static final Method PAGE_VIEW;
     private static final Method UPDATE_PAGE_PROPERTIES;
@@ -34,7 +35,7 @@ public class FullStoryModuleImpl {
             pageView = null;
             updatePageProperties = null;
             endPage = null;
-            Log.println(Log.ERROR, NAME, "Unable to access native FullStory pages API. Pages API will not function correctly. " +
+            Log.e(TAG, "Unable to access native FullStory pages API. Pages API will not function correctly. " +
                     "Make sure that your plugin is at least version 1.38; if the issue persists, please contact FullStory Support.");
         }
 
@@ -179,7 +180,7 @@ public class FullStoryModuleImpl {
             PAGE_VIEW.invoke(null, nonce, pageName, toMap(pageProperties));
         } catch (Throwable t) {
             // this should never happen
-            Log.println(Log.ERROR, NAME, "Unexpected error while calling startPage. Please contact FullStory Support.");
+            Log.e(TAG, "Unexpected error while calling startPage. Please contact FullStory Support.");
         }
     }
 
@@ -193,7 +194,7 @@ public class FullStoryModuleImpl {
             UPDATE_PAGE_PROPERTIES.invoke(null, nonce, toMap(pageProperties));
         } catch (Throwable t) {
             // this should never happen
-            Log.println(Log.ERROR, NAME, "Unexpected error while calling updatePage. Please contact FullStory Support.");
+            Log.e(TAG, "Unexpected error while calling updatePage. Please contact FullStory Support.");
         }
     }
 
@@ -206,7 +207,7 @@ public class FullStoryModuleImpl {
             END_PAGE.invoke(null, nonce);
         } catch (Throwable t) {
             // this should never happen
-            Log.println(Log.ERROR, NAME, "Unexpected error while calling endPage. Please contact FullStory Support.");
+            Log.e(TAG, "Unexpected error while calling endPage. Please contact FullStory Support.");
         }
     }
 }
--- a/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
+++ b/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
@@ -1,5 +1,7 @@
 package com.fullstory.reactnative;
 
+import androidx.annotation.NonNull;
+
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReadableMap;
@@ -9,9 +11,10 @@ import com.fullstory.FSPage;
 import com.fullstory.FSOnReadyListener;
 import com.fullstory.FSSessionData;
 
+import java.util.UUID;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
-
+import java.lang.reflect.Method;
 public class FullStoryModuleImpl {
 
     public static final String NAME = "FullStory";
@@ -140,25 +143,42 @@ public class FullStoryModuleImpl {
         return map.toHashMap();
     }
 
-    public static void createPage(String pageName, ReadableMap pageProperties) {
-        page = FS.page(pageName, toMap(pageProperties));
-    }
+    @NonNull
+    public static String getUUID() {
+		UUID uuid = UUID.randomUUID();
+        return uuid.toString();
+	}
+    
+    public static void startPage(String uuid, String pageName, ReadableMap pageProperties) {
+        UUID nonce = UUID.fromString(uuid);
 
-    public static void startPage(ReadableMap pageProperties) {
-        if (page != null) {
-            page.start(toMap(pageProperties));
+        try {
+            Method getWebViewClientMethod = FS.class.getMethod("__pageView", UUID.class, String.class, Map.class);
+            getWebViewClientMethod.invoke(null, nonce, pageName, toMap(pageProperties));
+        } catch (Throwable t) {
+            System.out.println("Unable to invoke __pageView via reflection, please contact FullStory Support.");
         }
     }
 
-    public static  void updatePage(ReadableMap pageProperties) {
-        if (page != null) {
-            page.updateProperties(toMap(pageProperties));
+    public static void updatePage(String uuid, ReadableMap pageProperties) {
+        UUID nonce = UUID.fromString(uuid);
+
+        try {
+            Method getWebViewClientMethod = FS.class.getMethod("__updatePageProperties", UUID.class, Map.class);
+            getWebViewClientMethod.invoke(null, nonce, toMap(pageProperties));
+        } catch (Throwable t) {
+            System.out.println("Unable to invoke __updatePageProperties, please contact FullStory Support. ");
         }
     }
 
-    public static void endPage() {
-        if (page != null) {
-            page.end();
+    public static void endPage(String uuid) {
+        UUID nonce = UUID.fromString(uuid);
+
+        try {
+            Method getWebViewClientMethod = FS.class.getMethod("__endPage", UUID.class);
+            getWebViewClientMethod.invoke(null, nonce);
+        } catch (Throwable t) {
+            System.out.println("Unable to invoke __endPage, please contact FullStory Support. ");
         }
     }
 }
--- a/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
+++ b/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
@@ -1,6 +1,5 @@
 package com.fullstory.reactnative;
 
-import androidx.annotation.NonNull;
 import android.util.Log;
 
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
```diff
@@ -140,25 +143,42 @@ private static Map toMap(ReadableMap map) {
...
+    public static String getUUID() {
+		UUID uuid = UUID.randomUUID();
+        return uuid.toString();
+	}
+    
+    public static void startPage(String uuid, String pageName, ReadableMap pageProperties) {
+        UUID nonce = UUID.fromString(uuid);
+
+        try {
+            Method getWebViewClientMethod = FS.class.getMethod("__pageView", UUID.class, String.class, Map.class);
```
ReeceLaF, line 156: The variable names of these methods should be updated.

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
```diff
@@ -140,25 +143,42 @@ private static Map toMap(ReadableMap map) {
...
+            getWebViewClientMethod.invoke(null, nonce, pageName, toMap(pageProperties));
+        } catch (Throwable t) {
+            System.out.println("Unable to invoke __pageView via reflection, please contact FullStory Support.");
```
ReeceLaF, line 159: This could happen if FS is disabled or if the FS version is too low. We should communicate this and probably not print the message every time. We should try to initialize method references once, the first time they are needed and cache them. We may need to detect the FS version and only attempt this if it is at least 1.38.

src/index.js
```diff
@@ -52,7 +51,6 @@ export default {
   log,
   resetIdleTimer,
   LogLevel,
-  createPage: createPageWithProperties,
```
ReeceLaF, line 55: `createPage`, which we should update to just `page` for consistency, should be the only page-related method exposed here.

src/FSPage.ts
```diff
@@ -0,0 +1,109 @@
...
+                    "called on the same FSPage instance that the corresponding `start` is " +
+                    "called on."
+            );
+            return;
+        }
+        endPage(this.nonce);
+        this.nonce = '';
+    }
+
+    getPageName() {
```
ReeceLaF, line 102: `getPageName` and `getProperties` are not needed here and should be removed.

src/FSPage.ts
```diff
@@ -0,0 +1,109 @@
...
+        // encounter a non-dictionary value
+        // (see com.fullstory.instrumentation FSPageImpl.java)
+        if (!FSPage.isObject(oldValue) || !FSPage.isObject(newValue)) {
+            return newValue;
+        }
+        return FSPage.mergeObjects(oldValue, newValue);
+    }
+
+    static mergeObjects(oldObj, newObj) {
+        if (!newObj) return oldObj;
```
ReeceLaF, line 44: This check isn't needed since both arguments are checked to be Objects before calling this method.

src/FSPage.ts
```diff
@@ -0,0 +1,109 @@
...
+                oldObj[key] = FSPage.merge(oldInnerValue, newObj[key]);
+            } else {
+                oldObj[key] = newObj[key];
+            }
+        }
+      
+        return oldObj;
+    }
+
+    cleanProperties() {
```
ReeceLaF, line 58: This and the other static methods and field should be private.

ReeceLaF (commented):
This is a step in the right direction for the current approach. I think there are some more ramifications / pitfalls we should talk about tomorrow though.

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
```diff
@@ -140,25 +143,42 @@ private static Map toMap(ReadableMap map) {
...
-    }
-
-    public static void startPage(ReadableMap pageProperties) {
-        if (page != null) {
-            page.start(toMap(pageProperties));
+    @NonNull
+    public static String getUUID() {
+		UUID uuid = UUID.randomUUID();
+        return uuid.toString();
+	}
```
jwise, line 150: Something strange has happened with the indentation here.

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
jwise: For customer visible errors, I generally try to phrase errors as "broad category, impact, action" (there is no need to expose internal implementation details if there is nothing they can do about it).  So I would write this:
```suggestion
            System.out.println("Unable to access native FullStory pages API.  Pages API will not function correctly.  Make sure that your plugin is at least version 1.38; if the issue persists, please contact FullStory Support.");
```

Also, is System.out.println() the correct logging API here?

src/index.js
RyanCommits: `createPage` will be replaced with `new FSPage('pageName')` as the new API.

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
RyanCommits: Using android.util.Log now

src/index.js
```diff
@@ -34,10 +32,11 @@ const LogLevel = {
   Assert: 5, // Clamps to Error on Android
 };
 
-const createPageWithProperties = (pageName, pageProperties = {}) => createPage(pageName, pageProperties);
-const startPageWithProperties = (pageProperties = {}) => startPage(pageProperties);
+const startPageWithProperties = (nonce, pageName, pageProperties = {}) => startPage(nonce, pageName, pageProperties);
 const identifyWithProperties = (uid, userVars = {}) => identify(uid, userVars);
 
+export { FSPage } from "./FSPage";
```
RyanCommits, line 38: New API will be: 
```
import FullStory, { FSPage } from '@fullstory/react-native';

const page = new FSPage('home page');
```

I think doing `new FullStory.FSPage('home page');` is too verbose for creating new class instances, so I'm exporting FSPage as its own named export.

src/NativeFullStory.ts
```diff
@@ -15,10 +15,10 @@ export interface Spec extends TurboModule {
   restart(): void;
   log(logLevel: number, message: string): void;
   resetIdleTimer(): void;
-  createPage(pageName: string, pageProperties?: Object): void;
-  startPage(pageProperties?: Object): void;
-  endPage(): void;
-  updatePage(pageProperties: Object): void;
+  startPage(nonce: string, pageName: string, pageProperties?: Object): void;
```
RyanCommits, line 18: This is to link the native methods to JavaScript methods for Turbo Modules (new architecture)

src/FSPage.ts
```diff
@@ -0,0 +1,100 @@
...
+            );
+            return;
+        }
+
+        this.properties = FSPage.merge(this.properties, properties);
+        this.cleanProperties();
+        updatePage(this.nonce, this.properties);
+    }
+
+    async start(properties?: Object) {
```
RyanCommits, line 79: `page.start()` will be async due to `getUUID` being async. The limitation here is that users cannot immediately call `page.end()` or `page.update()` immediately after `page.start()` until after `page.start()` is resolved successfully. 

 We're using native hardware to generate a cryptographically strong `nonce`. If the security of the `nonce` here is not important, we can generate a `nonce` through JavaScript using `Math.random()` and avoid async native code. @ReeceLaF what do you think?

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
```diff
@@ -1,21 +1,51 @@
...
+        Method endPage;
+        try {
+            pageView = FS.class.getMethod("__pageView", UUID.class, String.class, Map.class);
+            updatePageProperties = FS.class.getMethod("__updatePageProperties", UUID.class, Map.class);
+            endPage = FS.class.getMethod("__endPage", UUID.class);
+        } catch (Throwable t) {
+            pageView = null;
+            updatePageProperties = null;
+            endPage = null;
+            Log.println(Log.ERROR, NAME, "Unable to access native FullStory pages API. Pages API will not function correctly. " +
```
RyanCommits, line 38: It's not straightforward to detected SDK version number in this plugin. We either need to expose `BuildConfig` in the SDK or somehow pass in the version from gradle. I think this error does its job, but let me know your thoughts.

ReeceLaF (commented):
The async issue with the UUID is unfortunate, and we need to address that, but otherwise everything looks great!

src/FSPage.ts
RyanCommits: Added a Math.random() JS implementation of UUID

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
```diff
@@ -140,25 +170,47 @@ private static Map toMap(ReadableMap map) {
...
+    public static void startPage(String uuid, String pageName, ReadableMap pageProperties) {
+        if (!reflectionSuccess) {
+            return;
+        }
+
+        UUID nonce = UUID.fromString(uuid);
+        try {
+            PAGE_VIEW.invoke(null, nonce, pageName, toMap(pageProperties));
+        } catch (Throwable t) {
+            // this should never happen
```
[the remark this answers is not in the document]
RyanCommits: Added a message

src/FSPage.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.cleanProperties();
+    updatePage(this.nonce, this.properties);
+  }
+
+  async start(properties?: Object) {
+    if (properties) {
+      this.properties = FSPage.merge(this.properties, properties);
+      this.cleanProperties();
+    }
+    this.nonce = FSPage.generateUUID();
```
RyanCommits, line 86: Method is now synchronous

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
```diff
@@ -140,25 +169,44 @@ private static Map toMap(ReadableMap map) {
...
 
-    public static void startPage(ReadableMap pageProperties) {
-        if (page != null) {
-            page.start(toMap(pageProperties));
+        UUID nonce = UUID.fromString(uuid);
+        try {
+            PAGE_VIEW.invoke(null, nonce, pageName, toMap(pageProperties));
+        } catch (Throwable t) {
+            // this should never happen
+            Log.println(Log.ERROR, NAME, "Unexpected error while calling startPage. Please contact FullStory Support.");
```
ReeceLaF, line 182: Just a nit, but you can use `Log.e` instead of `Log.println(Log.ERROR, ...` to simplify the call. Also we use "fullstory" instead of "FullStory" for the tag name in the Android logging. It may be good to keep that consistent for case-sensitive filtering.

ReeceLaF (approved):
Changes LGTM with start being synchronous now. Found a nit, but nothing that would block. Great job with this!

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
RyanCommits: Set the TAG to `FullStoryModuleImpl` based on the the google [docs](https://developer.android.com/reference/android/util/Log) convention to use the class name. It'll also be less confusing since `fullstory` may communicate that the message is coming from the SDK.

ReeceLaF: That doc only says it's a good convention to define a tag in the class. When developing an app, it is good to use the class name, because it makes it easy to locate where the message is from, but because this is a library, we prefer to use the same tag to make it easy to filter and find / see messages coming from fullstory. Differentiating between the native SDK and the RN plugin is a good point though. Maybe we could do `"fullstoryRN"` so we would still get case-sensitive matching, but also differentiate the plugin logs.

RyanCommits: Heard. I'll make sure to make that change before the final merge on the feature branch

RyanCommits: Heard. I'll be sure to make that change on the feature branch itself.

PR #69, opened 2023-07-25
call private ios pages API

RyanCommits:
With the following JS code:
```
const startRecording = () => {
    let page;
    page = new FullStory.FSPage('Home', { tired: 'man', obj: { hmm: 'hmm' }, aaa: { hmm: 'hmm' } });
    page.start({ pageName: 'erp', Homes: 'erp', woman: 'wo', obj: 3, aaa: { pop: 'op' } });
    page.update({ updates: 'updates' });
    page.end();
  };

  const stopRecording = () => {
    let page;
    page = new FullStory.FSPage('Test');
    page.start();
    page.update({ updates: 'updates' });
    page.end();
  };
```
Old RN Architecture: 
https://app.staging.fullstory.com/ui/KWH/session/4688480919289856:4741929862889472

New RN Architecture:
https://app.staging.fullstory.com/ui/KWH/session/4688480919289856:6558104800395264

~~* Fix TS types~~
~~* Fix immutable object merge issue (Old Architecture)~~
~~* Fix Java method declaration on Old Architecture~~

### DIFF
--- a/android/src/legacy/java/com/FullStoryModule.java
+++ b/android/src/legacy/java/com/FullStoryModule.java
@@ -91,4 +91,9 @@ public class FullStoryModule extends ReactContextBaseJavaModule {
     public static void endPage(String uuid) {
         FullStoryModuleImpl.endPage(uuid);
     }
+
+    @ReactMethod
+    public static void getUUID(Promise promise) {
+        FullStoryModuleImpl.getUUID(promise);
+    }
 }
--- a/android/src/legacy/java/com/FullStoryModule.java
+++ b/android/src/legacy/java/com/FullStoryModule.java
@@ -91,9 +91,4 @@ public class FullStoryModule extends ReactContextBaseJavaModule {
     public static void endPage(String uuid) {
         FullStoryModuleImpl.endPage(uuid);
     }
-
-    @ReactMethod
-    public static void getUUID(Promise promise) {
-        FullStoryModuleImpl.getUUID(promise);
-    }
 }
--- a/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
+++ b/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
@@ -35,7 +35,7 @@ public class FullStoryModuleImpl {
             updatePageProperties = null;
             endPage = null;
             Log.println(Log.ERROR, NAME, "Unable to access native FullStory pages API. Pages API will not function correctly. " +
-                    "Make sure that your plugin is at least version 1.38; if the issue persists, please contact FullStory Support.");
+                    "Make sure that your plugin is at least version 1.41; if the issue persists, please contact FullStory Support.");
         }
 
         PAGE_VIEW = pageView;
--- a/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
+++ b/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
@@ -1,6 +1,7 @@
 package com.fullstory.reactnative;
 
 import androidx.annotation.NonNull;
+import android.util.Log;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
@@ -17,7 +18,6 @@ import java.lang.reflect.Method;
 public class FullStoryModuleImpl {
 
     public static final String NAME = "FullStory";
-    private static final String TAG = "FullStoryModuleImpl";
     public static final boolean reflectionSuccess;
     private static final Method PAGE_VIEW;
     private static final Method UPDATE_PAGE_PROPERTIES;
@@ -35,7 +35,7 @@ public class FullStoryModuleImpl {
             pageView = null;
             updatePageProperties = null;
             endPage = null;
-            Log.e(TAG, "Unable to access native FullStory pages API. Pages API will not function correctly. " +
+            Log.println(Log.ERROR, NAME, "Unable to access native FullStory pages API. Pages API will not function correctly. " +
                     "Make sure that your plugin is at least version 1.38; if the issue persists, please contact FullStory Support.");
         }
 
@@ -43,7 +43,7 @@ public class FullStoryModuleImpl {
         UPDATE_PAGE_PROPERTIES = updatePageProperties;
         END_PAGE = endPage;
 
-        reflectionSuccess = PAGE_VIEW != null && UPDATE_PAGE_PROPERTIES != null && END_PAGE != null;
+        reflectionSuccess = PAGE_VIEW  != null && UPDATE_PAGE_PROPERTIES != null && END_PAGE != null;
     }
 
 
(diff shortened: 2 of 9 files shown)


android/src/legacy/java/com/FullStoryModule.java
```diff
@@ -88,22 +88,22 @@ public static void resetIdleTimer() {
...
-        FullStoryModuleImpl.updatePage(pageProperties);
+    public static void endPage(String uuid) {
+        FullStoryModuleImpl.endPage(uuid);
     }
 
     @ReactMethod
-    public static void endPage() {
-        FullStoryModuleImpl.endPage();
+    public static String getUUID() {
+        return FullStoryModuleImpl.getUUID();
```
jwise, line 107: Are these intentionally in this PR?

ios/FullStory.mm
```diff
@@ -129,38 +129,31 @@ - (void) getCurrentSessionURL:(RCTPromiseResolveBlock)resolve reject:(RCTPromise
...
-RCT_EXPORT_METHOD(createPage:(NSString *)pageName pageProperties:(NSDictionary *)pageProperties)
+RCT_EXPORT_METHOD(startPage:(NSString *)nonce pageName:(NSString *)pageName pageProperties:(NSDictionary *)pageProperties)
 {
+    NSUUID *uuid = [[NSUUID alloc] initWithUUIDString:nonce];        
+
 	dispatch_async(dispatch_get_main_queue(), ^{
-		self->_page = [FS pageWithName:pageName properties:pageProperties];
+		[FS pageViewWithNonce:uuid name:pageName properties:pageProperties];
 	});
 }
```
jwise, line 139: check consistency of indentation

ios/FullStory.mm
```diff
@@ -129,38 +129,31 @@ - (void) getCurrentSessionURL:(RCTPromiseResolveBlock)resolve reject:(RCTPromise
...
 
-RCT_EXPORT_METHOD(createPage:(NSString *)pageName pageProperties:(NSDictionary *)pageProperties)
+RCT_EXPORT_METHOD(startPage:(NSString *)nonce pageName:(NSString *)pageName pageProperties:(NSDictionary *)pageProperties)
 {
+    NSUUID *uuid = [[NSUUID alloc] initWithUUIDString:nonce];        
+
 	dispatch_async(dispatch_get_main_queue(), ^{
-		self->_page = [FS pageWithName:pageName properties:pageProperties];
+		[FS pageViewWithNonce:uuid name:pageName properties:pageProperties];
```
jwise, line 137: [This comment in the related Android PR](https://github.com/fullstorydev/fullstory-react-native/pull/68/files#r1274392908) reminded me that this could be a problem if paired with a plugin that does not have `pageViewWithNonce`.  We presumably either need to check the plugin version -- or we need to check this with something like:
```suggestion
		if (![FS respondsToSelector:@selector(pageViewWithNonce:name:properties:)]) {
			...react native logging mechanism to complain...
		} else {
			[FS pageViewWithNonce:uuid name:pageName properties:pageProperties];
		}
```

ios/FullStory.mm
```diff
@@ -129,38 +129,31 @@ - (void) getCurrentSessionURL:(RCTPromiseResolveBlock)resolve reject:(RCTPromise
...
-}
+    NSUUID *uuid = [[NSUUID alloc] initWithUUIDString:nonce];
 
-RCT_EXPORT_METHOD(endPage)
-{
-	dispatch_async(dispatch_get_main_queue(), ^{
-		[self->_page end];
-	});
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [FS endPageWithNonce:uuid];
```
jwise, line 146: ditto

ios/FullStory.mm
```diff
@@ -129,38 +129,31 @@ - (void) getCurrentSessionURL:(RCTPromiseResolveBlock)resolve reject:(RCTPromise
...
-RCT_EXPORT_METHOD(updatePage:(NSDictionary *)pageProperties)
+RCT_EXPORT_METHOD(updatePage:(NSString *)nonce pageProperties:(NSDictionary *)pageProperties)
 {
-	dispatch_async(dispatch_get_main_queue(), ^{
-		[self->_page updateProperties:pageProperties];
-	});
+    NSUUID *uuid = [[NSUUID alloc] initWithUUIDString:nonce];        
+
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [FS updatePageWithNonce:uuid properties:pageProperties];
```
jwise, line 155: ditto

jwise (commented):
Hmm, Reece made a good point about compatibility with different versions.  Some comments inline.

ios/FullStory.mm
RyanCommits: added a check

ios/FullStory.mm
RyanCommits: Added a check

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
```diff
@@ -35,7 +35,7 @@ public class FullStoryModuleImpl {
             updatePageProperties = null;
             endPage = null;
             Log.println(Log.ERROR, NAME, "Unable to access native FullStory pages API. Pages API will not function correctly. " +
-                    "Make sure that your plugin is at least version 1.38; if the issue persists, please contact FullStory Support.");
+                    "Make sure that your plugin is at least version 1.41; if the issue persists, please contact FullStory Support.");
```
RyanCommits, line 38: Bumped this Android message to `1.41` since we edited the `iOS` private API's made available in `1.41`. We want the recommended version for Pages API to be consistent.

ReeceLaF (commented):
LGTM, but I'll let @jwise approve as its iOS.

android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
martin-fs: This check and informative log message is Android-specific. We should have a similar check and log message for iOS.

RyanCommits: The error message `PagesAPIError` in `ios/FullStory.mm` serves this purpose. Not sure if you missed it or if you mean something else?

ios/FullStory.mm
```diff
@@ -129,38 +131,43 @@ - (void) getCurrentSessionURL:(RCTPromiseResolveBlock)resolve reject:(RCTPromise
...
 
-RCT_EXPORT_METHOD(endPage)
+RCT_EXPORT_METHOD(updatePage:(NSString *)nonce pageProperties:(NSDictionary *)pageProperties)
 {
-	dispatch_async(dispatch_get_main_queue(), ^{
-		[self->_page end];
-	});
-}
+	if (![FS respondsToSelector:@selector(_updatePageWithNonce:properties:)]) {
+		RCTLogError(PagesAPIError, @"endPage");
```
martin-fs, line 163: I think "endPage" should be "updatePage" here.

ios/FullStory.mm
```diff
@@ -5,14 +5,16 @@
...
+#import <React/RCTLog.h>
 
 #import "FSReactSwizzle.h"
 
 @implementation FullStory {
 	RCTPromiseResolveBlock onReadyPromise;
-	id<FSPage> _page;
 }
 
+NSString *const PagesAPIError = @"Unable to call %@. Make sure that your plugin is at least version 1.41; if the issue persists, please contact FullStory Support.";
```
martin-fs, line 16: Could you add some verbiage similar to the Android log, like `"Unable to access native FullStory pages API. Pages API will not function correctly. "`. This will help the clarity of the log message and the consistency between platforms.

martin-fs (approved):
Thanks for pointing out that iOS log message - I think it's because I was skimming for an "Unable to access pages" type of message and didn't see it. So, consider a couple of small string changes, but otherwise looks good!

PR #70, opened 2023-07-28
update Android gradle configs

RyanCommits:
* Remove Android support lib
* Remove gradle tools version

Followed RN community standards with these configurations. Tested on RN `0.63.5`, `0.72.0` new & old architecture, `0.69.9`

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -88,6 +88,8 @@ if (match.matches()) {
 }
 
 dependencies {
+    // For < 0.71, this will be from the local maven repo
+    // For > 0.71, this will be replaced by `com.facebook.react:react-android:$version` by react gradle plugin
     //noinspection GradleDynamicVersion
     implementation 'com.facebook.react:react-native:+'
     implementation "com.fullstory:instrumentation-full:$fsDetectedVersion@aar"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,20 +2,6 @@ def safeExtGet(prop, fallback) {
     rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
 }
 
-buildscript {
-    /* In case of submodule usage, do not try to apply own repositories and plugins,
-        root project is responsible for that. */
-    if (rootProject.buildDir == project.buildDir) {
-        repositories {
-            google()
-            jcenter()
-        }
-        dependencies {
-            classpath 'com.android.tools.build:gradle:3.4.1'
-        }
-    }
-}
-
 def isNewArchitectureEnabled() {
   return rootProject.hasProperty("newArchEnabled") &&  rootProject.getProperty("newArchEnabled") == "true"
 }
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -102,7 +88,7 @@ if (match.matches()) {
 }
 
 dependencies {
-    implementation "com.android.support:support-v4:${safeExtGet('supportLibVersion', '27.+')}"
-    api "com.facebook.react:react-native:${safeExtGet('reactNative', '+')}"
+    //noinspection GradleDynamicVersion
+    implementation 'com.facebook.react:react-native:+'
```
ReeceLaF, line 92: Changing to implementation is probably the correct thing to do here, but why change the version to not match what is already in the project?

android/build.gradle
```diff
@@ -2,20 +2,6 @@ def safeExtGet(prop, fallback) {
     rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
 }
 
-buildscript {
```
ReeceLaF, line 5: This removal makes me a bit nervous. Are you sure this wasn't needed for any reason?

ReeceLaF: @prashanthsadasivan-fullstory Do you have any context on why this was here?

android/build.gradle
RyanCommits: This is from the current react-native library template 

https://github.com/callstack/react-native-builder-bob/blob/44df3f943f7c96bb022150348cfdfaa79a8fdca0/packages/create-react-native-library/templates/native-common/android/build.gradle#L202

android/build.gradle
RyanCommits: This `if` statement is never true, since `rootProject.buildDir` && `project.buildDir` would not be equal, since this plugin would always be a submodule of the root project.

android/build.gradle
RyanCommits: Examples of public libraries without this buildscript:

https://github.com/software-mansion-labs/react-native-webrtc/blob/master/android/build.gradle
https://github.com/getsentry/sentry-react-native/blob/main/android/build.gradle

android/build.gradle
RyanCommits: Examples of public libraries without this buildscript:

https://github.com/software-mansion-labs/react-native-webrtc/blob/master/android/build.gradle
https://github.com/getsentry/sentry-react-native/blob/main/android/build.gradle

ReeceLaF: _Is_ it always a subproject of the root gradle project? Other projects not having buildscript blocks doesn't mean this one shouldn't have it, but this one doesn't seem to be doing anything important. As long as you've verified this change doesn't break anything, I'm ok with taking this out.

android/build.gradle
ReeceLaF: Let's copy the comments about the behavior over too.

android/build.gradle
RyanCommits: Since this is a react-native plugin, this gradle file will always be run from a root project's `node_modules` directory. The root project will always contain its own gradle configuration. 

This project also could never be built on its own, since the way we read `$fsDetectedVersion` is by reading the root project's `'gradle-plugin-local-'`, which will never exist in this gradle file.

IMO it's safe to delete this and avoid having to repeatedly update the build tools version that will never get read.

PR #75, opened 2023-08-29
tsconfig setup, type fix, and eslint build step

RyanCommits:
I'll describe what I did in the order of the four commits in this PR:

1. Set up a tsconfig and prepack script so we can use `tsc` to generate `.d.ts` type files. Move `index.js` to `index.ts`. Transpiled JS code will live in `lib/`. We wrote `index.d.ts` by hand, but we'll use the ts transpiler instead in the future.
2. Fix `FS.identify` type because 2nd argument is not optional, as it will crash in iOS with only 1 argument.
3. Fix eslint warnings by fixing variable shadowing. No functional changes.
4. Add linting to build step


src/index.ts
```diff
@@ -19,10 +37,16 @@ interface UserVars {
...
+  replayStartUrl: string;
+  replayNowUrl: string;
+  sessionId: string;
+};
+
 declare type FullStoryStatic = {
   LogLevel: typeof LogLevel;
   anonymize(): void;
-  identify(uid: string, userVars?: UserVars): void;
+  identify(uid: string, userVars: UserVars): void;
```
JoshMiers-FS, line 49: will this change impact existing code that doesn't pass in a second param?

src/index.ts
```diff
@@ -45,5 +69,20 @@ declare global {
...
+  consent,
+  event,
+  shutdown,
+  restart,
+  log,
+  resetIdleTimer,
+  LogLevel,
+};
+
+export default FullStoryAPI;
```
JoshMiers-FS, line 88: how does exporting FullStoryAPI instead of FullStory impact things?

src/index.ts
RyanCommits: The existing code that doesn't pass a second param will show a TypeScript warning if they're using TypeScript, but functionally it won't change anything.

src/index.ts
RyanCommits: The `index.d.ts` that's generated will have:
```
declare const FullStoryAPI: FullStoryStatic;
export default FullStoryAPI;
```

The renaming of `FullStory` -> `FullStoryAPI` has no functional impact. Users will still use `import FullStory from '@fullstory/react-native';`

src/index.ts
JoshMiers-FS: Is there a way to allow for them to call identify(string) and identify(string, UserVars) without a warning? We support that in the regular Android API so I am concerned about generating a warning for a valid use of the API.

RyanCommits: Fixed

JoshMiers-FS (approved):
LGTM

src/index.ts
martin-fs: If a customer does `import * from '@fullstory/react-native'`, will they now get a `const FullStoryAPI` instead of a `const FullStory` object?

martin-fs: Sorry for the noise here. I agree, it should do exactly the same thing - `export default X` exports `X` as its "default" object, and since `FullStory` / `FullStoryAPI` are not exported by name, `import *` will do nothing. [This was the clearest explanation I could find on this whole situation](https://stackoverflow.com/a/33307487) (under "Additional Points").

PR #78, opened 2023-10-10
Auto capture console logs

RyanCommits:
Issue: https://fullstory.atlassian.net/browse/MOCA-8076

Add methods `enableConsole ` and `disableConsole` to FullStory API that will extend React Native's `console` API. Messages will be passed to `FS.log`. 

Console capture defaults to enabled, and follows the FS default log level of `info` if the customer hasn't configured it already.

Example session: https://app.staging.fullstory.com/ui/KWH/session/6727898337050624:5966357937455104/devtools/console

Added enable and disable API.

Added `safeStringify` from web side, minus `DOM` node handling since that's not defined in RN.

### DIFF
--- a/src/__tests__/safeStringify.test.ts
+++ b/src/__tests__/safeStringify.test.ts
@@ -1,0 +1,174 @@
+import { safeStringify } from '../logging/safeStringify';
+
+const LARGE_MAX_CHARS = 10000; // Arbitrarily large.
+
+const expectValidJSON = (str: string) => expect(() => JSON.parse(str)).not.toThrowError();
+
+describe('safeStringify', () => {
+  it('handles character limits', () => {
+    const a1 = [1, 2, 3, [4, 5, [6, 7, 8], 9], 10];
+
+    {
+      // No truncation
+      const act = safeStringify(a1, LARGE_MAX_CHARS);
+      expect(act).toBe('[1,2,3,[4,5,[6,7,8],9],10]');
+      expectValidJSON(act);
+    }
+
+    {
+      // truncate an array
+      const act = safeStringify(a1, 10);
+      expect(act).toBe('[1,2,3,[]]');
+      expectValidJSON(act);
+    }
+
+    {
+      // truncate on nested array
+      const act = safeStringify(a1, 16);
+      expect(act).toBe('[1,2,3,[4,5,[]]]');
+      expectValidJSON(act);
+    }
+
+    {
+      // truncate really short array (no trailing comma!)
+      const act = safeStringify(a1, 4); // Test for trailing comma
+      expect(act).toBe('[1]');
+      expectValidJSON(act);
+    }
+
+    {
+      // no truncate regular object
+      const o1 = { a: 1, b: 2, c: null, d: undefined, e: 'asdf' };
+      const act = safeStringify(o1, LARGE_MAX_CHARS);
+      expect(act).toBe('{"a":1,"b":2,"c":null,"d":"undefined","e":"asdf"}');
+      expectValidJSON(act);
+    }
+
+    const o2 = { a: 1, b: 2, c: null, d: [1, 2, ['three'], [], {}, []], e: 'asdf' };
+
+    {
+      // truncation on regular object
+      const act = safeStringify(o2, 13);
+      expect(act).toBe('{"a":1,"b":2}');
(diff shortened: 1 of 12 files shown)


RyanCommits:
Updated jest dev dependencies and snapshots to resolve npm errors.

src/index.ts
```diff
@@ -73,6 +74,35 @@ const identifyWithProperties = (uid: string, userVars = {}) => identify(uid, use
 
 export { FSPage } from './FSPage';
 
+const captureConsoleLogs = () => {
+  const consoleLevelMap = {
```
RyanCommits, line 78: Just thought of an idea - We'll instruct the user to only run this once, but we can wrap this function in a closure so that if they run this multiple times by mistake, we'll skip over the function body.

src/index.ts
```diff
@@ -73,6 +74,35 @@ const identifyWithProperties = (uid: string, userVars = {}) => identify(uid, use
 
 export { FSPage } from './FSPage';
 
+const captureConsoleLogs = () => {
```
prashanthsadasivan-fullstory, line 77: Should the customer be able to configure this so that they can specifiy what level they care about? I can see some customers wanting to omit a debug logs, but keep error logs.

prashanthsadasivan-fullstory (commented):
One comment about configuration

src/index.ts
RyanCommits: This function will eventually call `FS.log`, which is controlled by `LogLevel` native config. It may be confusing if we require users to configure `LogLevel` as well as passing in the right `logLevel` to this JS function. We also don't have this configuration on the web side, and arguably, the workflows that React Native developers have are similar to web developers. What do you think?

prashanthsadasivan-fullstory: @RyanCommits hmm, I think that would only work for Android no? I don't think iOS has a native `LogLevel` config right?  I'm okay with it being something to decide in a followup PR though

RyanCommits: https://github.com/cowpaths/mn/commit/9910bc458a346c4ed982a4fe479a7d818ad0ff03

We have a LogLevel config on iOS. I don't believe it's documented anywhere

RyanCommits:
After discussion - we discovered new requirements: https://docs.google.com/document/d/1QAGpjrC96ePbY_-SugOYK9zFSaqh1HnkXh6CP5MrRp8

src/logging/consoleCapture.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  for (let i = 0; i < args.length; ++i) {
+    payload.push(safeStringify(args[i], 1000));
+  }
+
+  FullStory.log(level, payload.join(' '));
+};
+
+// A naive version of ConsoleWatcher on web.
+// See https://github.com/cowpaths/mn/blob/865353f374b687e481d3b230e7eec8e1d7be2eb4/projects/fullstory/packages/recording/src/consolewatcher.ts
```
jurassix, line 47: this is a public repo correct? doesn't seem that useful to ref our private repo

src/logging/consoleCapture.ts
```diff
@@ -0,0 +1,104 @@
+import { NativeModules } from 'react-native';
+import { safeStringify } from './safeStringify';
+
+// @ts-expect-error
+const isTurboModuleEnabled = global.__turboModuleProxy != null;
```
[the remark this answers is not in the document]
jurassix: I've added this to my pr feedback branch.

src/__tests__/safeStringify.test.ts
```diff
@@ -0,0 +1,174 @@
+import { safeStringify } from '../logging/safeStringify';
```
jurassix, line 1: since this is a public repo seems like we could add a third-party dep for this instead of managing this directly.

jurassix (commented):
Overall seems fine. Added my nit feedback to a stacked pr: https://github.com/fullstorydev/fullstory-react-native/pull/83

Two open issues:
- do we need to expose the safeStringify or can we use an OSS solution? Longer term if we want to maintain this functinality it seems best to do as a separart lib.
- It's unclear why we default to enable by default and not let the FSTA settings determine the initial state.

RyanCommits:
Closing this for now since we'll need to discuss console scrubbing on the native side

PR #79, opened 2023-10-31
iOS Fabric implementation

RyanCommits:
See [design doc](https://docs.google.com/document/d/1OIahWaovsaoqdWr12zmJRcF0LZlOzoDuLcv6FPzz_WY)

This work seeks to solve an issue where props are statically defined in Fabric and cannot set our FullStory props onto views.

### DIFF
--- a/ios/FullStory.mm
+++ b/ios/FullStory.mm
@@ -216,6 +216,38 @@ RCT_REMAP_METHOD(onReady, onReadyWithResolver:(RCTPromiseResolveBlock)resolve re
 
 static const char *_rctview_previous_attributes_key = "associated_object_rctview_previous_attributes_key";
 
+static void set_fsClass(id json, RCTView *view) {
+    NSArray <NSString *>* classes = [(NSString *)json componentsSeparatedByString: @","];
+    [FS removeAllClasses:view];
+    [FS addClasses:view classNames:classes];
+}
+
+
+static void set_fsAttribute(id json, RCTView *view) {
+    NSDictionary *newAttrs = (NSDictionary *)json;
+
+    /* Clear up all the old attributes first, if they exist. */
+    NSSet *oldAttrs = objc_getAssociatedObject(view, _rctview_previous_attributes_key);
+    if (oldAttrs) {
+        for (NSString *attr in oldAttrs) {
+            [FS removeAttribute:view attributeName:attr];
+        }
+    }
+
+    /* Load in the new attributes. */
+    NSMutableSet *newAttrSet = [NSMutableSet new];
+    if (newAttrs) {
+        for (NSString *attr in newAttrs) {
+            [FS setAttribute:view attributeName:attr attributeValue:(NSString *)newAttrs[attr]];
+            [newAttrSet addObject:attr];
+        }
+    }
+
+    /* And set them up for cleanup next time. */
+    objc_setAssociatedObject(view, _rctview_previous_attributes_key, newAttrSet, OBJC_ASSOCIATION_RETAIN);
+}
+
+
 @implementation RCTViewManager (FullStory)
 
 + (NSArray<NSString *> *) propConfig_fsClass {
@@ -223,9 +255,7 @@ static const char *_rctview_previous_attributes_key = "associated_object_rctview
 }
 
 - (void) set_fsClass:(id)json forView:(RCTView*)view withDefaultView:(RCTView *)defaultView {
-	NSArray <NSString *>* classes = [(NSString *)json componentsSeparatedByString: @","];
-	[FS removeAllClasses:view];
-	[FS addClasses:view classNames:classes];
+    set_fsClass(json, view);
 }
 
 + (NSArray<NSString *> *) propConfig_dataComponent {
@@ -265,27 +295,7 @@ static const char *_rctview_previous_attributes_key = "associated_object_rctview
 }
 
 - (void) set_fsAttribute:(id)json forView:(RCTView*)view withDefaultView:(RCTView *)defaultView {
-	NSDictionary *newAttrs = (NSDictionary *)json;
-	
-	/* Clear up all the old attributes first, if they exist. */
-	NSSet *oldAttrs = objc_getAssociatedObject(view, _rctview_previous_attributes_key);
-	if (oldAttrs) {
-		for (NSString *attr in oldAttrs) {
-			[FS removeAttribute:view attributeName:attr];
-		}
-	}
-	
-	/* Load in the new attributes. */
-	NSMutableSet *newAttrSet = [NSMutableSet new];
-	if (newAttrs) {
-		for (NSString *attr in newAttrs) {
-			[FS setAttribute:view attributeName:attr attributeValue:(NSString *)newAttrs[attr]];
-			[newAttrSet addObject:attr];
-		}
-	}
-	
-	/* And set them up for cleanup next time. */
-	objc_setAssociatedObject(view, _rctview_previous_attributes_key, newAttrSet, OBJC_ASSOCIATION_RETAIN);
+    set_fsAttribute(json, view);
 }
 @end
 
@@ -309,5 +319,13 @@ static const char *_rctview_previous_attributes_key = "associated_object_rctview
 		r[@"propTypes"][@"fsAttribute"] = @"NSDictionary *";
 		return r;
 	} SWIZZLE_END;
+
+    SWIZZLE_BEGIN_INSTANCE(RCTViewComponentView, @selector(handleCommand:args:), void, const NSString *commandName, const NSArray *args) {
+            if ([commandName isEqual:@"fsAttribute"]) {
+                set_fsAttribute(args[0], self);
+            } else if ([commandName isEqualToString:@"fsClass"]) {
+                set_fsClass(args[0], self);
+            }
+        } SWIZZLE_END;
 }
 @end
--- a/ios/FullStory.mm
+++ b/ios/FullStory.mm
@@ -316,6 +316,32 @@ static void set_fsAttribute(id json, RCTView *view) {
 
 @interface FSReactSwizzleBootstrap : NSObject
 @end
+#define SWIZZLE_HANDLE_COMMAND(rct_clazz) \
+SWIZZLE_BEGIN_INSTANCE(rct_clazz, @selector(handleCommand:args:), void, const NSString *commandName, const NSArray *args) { \
+    if ([commandName isEqualToString:@"fsAttribute"]) { \
+        set_fsAttribute(args[0], self); \
+    } else if ([commandName isEqualToString:@"fsClass"]) { \
+        set_fsClass(args[0], self); \
+    } else if ([commandName isEqualToString:@"dataElement"]) { \
+        set_dataElement(args[0], self); \
+    } else if ([commandName isEqualToString:@"dataSourceFile"]) { \
+        set_dataSourceFile(args[0], self); \
+    } else if ([commandName isEqualToString:@"fsTagName"]) { \
+        set_fsTagName(args[0], self); \
+    } else if ([commandName isEqualToString:@"dataComponent"]) { \
+        set_dataComponent(args[0], self); \
+    } else { \
+        SWIZZLED_METHOD(commandName, args); \
+    } \
+} SWIZZLE_END
+
+static bool array_contains_string(const char **array, const char *string) {
+    if (string == nullptr || array == nullptr) return false;
+    for (; *array != nullptr; array++) {
+        if (strcmp(*array, string) == 0) return true;
+    }
+    return false;
+}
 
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -72,6 +84,74 @@ declare global {
...
+const Commands: NativeCommands = codegenNativeCommands<NativeCommands>({
+  supportedCommands: [
+    'fsClass',
+    'fsAttribute',
+    'fsTagName',
+    'dataElement',
+    'dataComponent',
+    'dataSourceFile',
+  ],
+});
```
jwise, line 107: As I mentioned in the design doc, I'm still a little nervous about doing these sequentially.  Do we know what thread JavaScript runs on?  I guess it might be okay iff a view is added to the hierarchy only after our `applyFSPropertiesWithRef` completes.  But if that's the case we need to document that, ideally, in a way that shows why that is the case.

src/index.ts
```diff
@@ -72,6 +84,74 @@ declare global {
...
+
+export function applyFSPropertiesWithRef(existingRef?: ForwardedRef<unknown>) {
+  return function (element: React.ElementRef<FSComponentType>) {
+    // https://github.com/facebook/react-native/blob/87d2ea9c364c7ea393d11718c195dfe580c916ef/packages/react-native/Libraries/Components/TextInput/TextInputState.js#L109C23-L109C67
+    // @ts-expect-error `currentProps` is missing in `NativeMethods`
+    if (element && element.currentProps) {
+      // @ts-expect-error `currentProps` is missing in `NativeMethods`
+      const fsClass = element.currentProps['fsClass'];
```
jwise, line 115: I wonder if we could DRY this `currentProps is missing` thing by pulling `currentProps` into a local.  And broadly I wonder if we could DRY it by just iterating all of the `supportedCommands` or something rather than naming each one three times...

jwise (changes requested):
I see one 'in review' on DOC-1863, one 'unstarted', and one 'design approved'.  I still have some open questions from a few weeks ago, and I think @jth0024 does as well, that need to be answered before we merge this approach (probably we should add Jordan as a reviewer on DOC-1863, also).  I think it is possible that we can make this design work, but I'm not sure we're ready to commit to it based on the state of the design doc?

So I guess -1 for now on that front (and certainly there are style / functionality changes that are currently needed), but I don't fundamentally object to this design if we can show that it works well.

RyanCommits:
Sorry @jwise , I should've labeled this PR as work in progress. I wanted to get my `JS` changes up so that @martin-fs  can fill in his `iOS` work after. Martin has the latest Obj-c code which isn't in this PR yet

ios/FullStory.mm
```diff
@@ -216,16 +216,61 @@ - (void)log:(double)logLevel message:(NSString *)message {
 
 static const char *_rctview_previous_attributes_key = "associated_object_rctview_previous_attributes_key";
 
+static void set_fsClass(id json, RCTView *view) {
```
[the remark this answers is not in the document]
martin-fs: I think a static function is the correct idiom to use when the functionality's use is limited to the current file. Extensions are meant to allow use across files - they add functions to the lookup table, and carry the risk of being referenced outside of their intended use.

ios/FullStory.mm
```diff
@@ -216,16 +216,61 @@ - (void)log:(double)logLevel message:(NSString *)message {
...
+    NSArray <NSString *>* classes = [(NSString *)json componentsSeparatedByString: @","];
```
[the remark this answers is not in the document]
martin-fs: I'm following the original code here to have minimal impact:
https://github.com/fullstorydev/fullstory-react-native/blob/45e21f3a7350c624221bfe9ab75639f3ef03ea9d/ios/FullStory.mm#L225-L228
The code has been in place for [over three years](https://github.com/fullstorydev/fullstory-react-native/pull/12) without issue.

It should not be possible with TS to include a non-string property in the Old Architecture:
https://github.com/fullstorydev/fullstory-react-native/blob/45e21f3a7350c624221bfe9ab75639f3ef03ea9d/src/index.ts#L62-L70
See the changes to `index.ts` below for the equivalent restriction in Fabric:
https://github.com/fullstorydev/fullstory-react-native/pull/79/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R90-R95

It seems appropriate to crash if a programmer attempts to subvert the type system. However, I could add a test here, and terminate FullStory if it is of the incorrect type, because continuing to capture is inappropriate if there is an attempt to set a privacy rule that cannot be handled. We would have to expose an API to terminate FullStory. This seems beyond the scope of this PR, but I would be happy to add it in a followup.

ios/FullStory.mm
```diff
@@ -216,16 +216,61 @@ - (void)log:(double)logLevel message:(NSString *)message {
...
+    [FS removeAllClasses:view];
+    [FS addClasses:view classNames:classes];
+}
+
+static void set_fsTagName(id json, RCTView *view) {
+   [FS setTagName:view tagName:(NSString *)json];
+}
```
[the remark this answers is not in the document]
martin-fs: I'm not sure what you're referring to when you say `fs_cast` here - there is no reference to it in this repository. Perhaps this is an example of an extension being referenced outside of its intended use. [isKindOfClass](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418511-iskindofclass) is public API that can be used in the test I proposed above. Otherwise, crashing would be better than continuing to capture after ignoring a privacy rule, and seems like an appropriate response to an attempt to subvert the type system.

src/index.ts
martin-fs: I believe your concerns were addressed in the discussion and design doc. I don't think it's appropriate to write an essay about React Native's architecture here.

src/index.ts
martin-fs: I looked at this and as far as I can tell, [it's not completely trivial](https://stackoverflow.com/a/45670960). Interface properties are not iterable and only exist at compile time unless customers implement custom transformers. Plugin support for custom transformers continues to be an [open issue after five years](https://github.com/Microsoft/TypeScript/issues/14419). This aesthetic issue should not be a blocking concern for this PR.

prashanthsadasivan-fullstory (approved):
the swizzles look good to me. some various small comments, but LGTM from my perspective.

ios/FullStory.mm
```diff
@@ -309,5 +360,70 @@ + (void) load {
...
+                // from the swizzle above.
+                // Since we skip the explicitly swizzled classes, each with
+                // its own implementation, the remaining classes should either
+                // have the superclass implementation, or their own, which is
+                // incapable of receiving FS properties.
+                Method _Nullable existingMethod = class_getInstanceMethod(cls, sel);
+                if (existingMethod) {
+                    IMP existingImplementation = method_getImplementation(existingMethod);
+                    if (existingImplementation != swizzledViewComponentViewCommandImplementation) {
+                        NSAssert(strncmp(className, "RCT", 3) != 0, @"React Native framework class %s needs handleCommand support! Please contact FullStory support with this message.", className);
```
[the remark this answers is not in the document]
martin-fs: Yes, it would crash the app when it is running in debug mode. I agree that this would be very disruptive to the developer. However, it's kind of unsafe for them to use the SDK if this is the case. If we are worried that this assert will be triggered, we could instead swizzle all subclasses per the original design. Or, we could add a terminate-with-error-message method to the SDK in a followup PR and terminate with this error message instead of NSAssert.

ios/FullStory.mm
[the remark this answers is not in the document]
jwise: Yes, I am OK with this if this is following the original.

src/index.ts
jwise: Yes, my concerns about threading of this were addressed in the design doc.  It might be worth adding a comment to, if nothing else, point to the discussion in the design doc for why this is ok.

src/index.ts
jwise: Wow, big oof that we can't iterate over an interface.

I was kind of thinking something like (pseudocode):

```javascript
  for (prop in ['fsClass', 'fsAttribute', 'fsTagName', ...]) {
    // @ts-expect-error ...
    const val = element.currentProps[prop];
    if (val) {
      Commands[prop](element, val); /* will we need a ts waiver here? */
    }
  }
```

Something like this definitely has the chance to grow copy-and-paste errors if it needs to be touched later so if we can avoid repeating something six times, that might be nice.  If we were feeling *extremely* fancy, that list (`['fsClass', ...]`) could also be a `const` elsewhere that we could pass as the parameter to `codegenNativeCommands`, for one more repetition eliminated!

Or, if nothing else, hoisting out `const props = element.currentProps;`, so we only have to `@ts-expect-error` once.  What do you think about these?

ios/FullStory.mm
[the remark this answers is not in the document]
jwise: Ok, fair enough.  (Generally I tend to prefix static methods with `_`, and I tend to prefer the parameter order `object_verb_noun(object, parameters...)`, which in this case would translate to `_rctview_set_fsClass(RCTView *view, id json)`, just to scope things a little more tightly -- like we do two lines prior, scoping `_rctview_previous_attributes_key` by name -- but that could be a matter of taste.)

ios/FullStory.mm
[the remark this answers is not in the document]
jwise: I'm ok with this being a crashing NSAssert -- especially because a developer wanting to work around it doesn't need to wait for us to release a new plugin version, but instead can make the modification themselves in their local tree.  Nice work on checking this in debug.

ios/FullStory.mm
```diff
@@ -309,5 +360,70 @@ + (void) load {
...
+                        NSLog(@"RCTViewComponentView subclass %s cannot receive FullStory commands", className);
```
jwise, line 421:
```suggestion
                        NSLog(@"RCTViewComponentView subclass %s cannot receive FullStory commands; FullStory attributes on such views may not function correctly.  If you need to attach FullStory attributes to such views, please contact FullStory support with this message.", className);
```

ios/FullStory.mm
```diff
@@ -309,5 +360,70 @@ + (void) load {
...
+    SWIZZLE_HANDLE_COMMAND(RCTViewComponentView);
+    SWIZZLE_HANDLE_COMMAND(RCTTextInputComponentView);
+    SWIZZLE_HANDLE_COMMAND(RCTSwitchComponentView);
+    SWIZZLE_HANDLE_COMMAND(RCTScrollViewComponentView);
+    SWIZZLE_HANDLE_COMMAND(RCTPullToRefreshViewComponentView);
+    SWIZZLE_HANDLE_COMMAND(RCTLegacyViewManagerInteropComponentView);
+#pragma clang pop
+#ifdef DEBUG
+    dispatch_async(dispatch_get_global_queue(QOS_CLASS_UTILITY, 0), ^{
+#pragma clang diagnostic push
```
jwise, line 377:
```suggestion
      // RCTViewComponentView subclasses don't tend to call their superclass implementations of handleCommand, so in debug mode, we want to make sure that we don't have any "straggler" classes (especially React core classes!) that don't have their implementations of handleCommand interposed.  We enumerate all the classes on the system, determine if any has a superclass of RCTViewComponentView, and if so, make sure that we have it covered.
#pragma clang diagnostic push
```

ios/FullStory.mm
```diff
@@ -309,5 +360,70 @@ + (void) load {
...
+        Class viewComponentView = NSClassFromString(@"RCTViewComponentView");
+        Method _Nullable swizzledViewComponentViewCommand = class_getInstanceMethod(viewComponentView, sel);
+        IMP swizzledViewComponentViewCommandImplementation = method_getImplementation(swizzledViewComponentViewCommand);
+
+        int classCount = objc_getClassList(NULL, 0);
+        // The classes stored in this array may not be NSObject classes,
+        // so may not respond to retain messages. See:
+        // https://gist.github.com/mikeash/1267596
+        __unsafe_unretained Class classes[classCount];
+        // Install swizzle in all subclasses of RCTViewComponentView
```
jwise, line 399:
```suggestion
        // Verify swizzle in all subclasses of RCTViewComponentView
```

jwise (approved):
ObjC changes LGTM.  I suggested a few comments and a few style changes, and it would be nice to see something to shore up the DRY situation down there in TS land, but I trust you to make those changes as appropriate.  Looks like an approach that is going to work out!

src/index.ts
RyanCommits: Hoisted the `element.currentProps` so we only have the `ts-expect-error` once. 

Improved types so copy-and-paste errors will be unlikely since it has to match the `NativeCommand` type.

src/index.ts
RyanCommits: comment added

PR #84, opened 2024-01-24
iOS: Use preprocessor variable for handleCommand check

martin-fs:


ios/FullStory.mm
```diff
@@ -372,59 +372,65 @@ + (void) load {
     SWIZZLE_HANDLE_COMMAND(RCTPullToRefreshViewComponentView);
     SWIZZLE_HANDLE_COMMAND(RCTLegacyViewManagerInteropComponentView);
 #pragma clang pop
-#ifdef DEBUG
-    dispatch_async(dispatch_get_global_queue(QOS_CLASS_UTILITY, 0), ^{
-      // RCTViewComponentView subclasses don't tend to call their superclass implementations of handleCommand, so in debug mode, we want to make sure that we don't have any "straggler" classes (especially React core classes!) that don't have their implementations of handleCommand interposed.  We enumerate all the classes on the system, determine if any has a superclass of RCTViewComponentView, and if so, make sure that we have it covered.
+#if defined(DEBUG_FS_RN_FABRIC_THIRD_PARTY)
```
martin-fs, line 375: We could also guard this behind `DEBUG` if we want to prevent running this code in release builds.

RyanCommits (approved):
lgtm

PR #85, opened 2024-02-13
Raise audit level to pass CI

RyanCommits:
Dependency [react-native-cli](https://github.com/react-native-community/cli/issues/2294) `ip` received a security warning which the community is in the process of resolving. We're raising the audit level here to prevent us from being blocked from making updates.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
           key: v1-dependencies-{{ checksum "package.json" }}
       - run:
           name: Audit
-          command: npm audit --audit-level=low
+          command: npm audit --audit-level=critical
       - run:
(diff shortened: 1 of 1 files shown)


martin-fs (approved):
I notice that even with this change, [we still see the issue in the build logs](https://app.circleci.com/pipelines/github/fullstorydev/fullstory-react-native/261/workflows/52af791d-707f-4519-8ffd-f181eee17d4d/jobs/268) so we can keep tabs on it, and we will still fail our build if a critical vulnerability shows up in the meantime. Reading the documentation on `npm audit`, there does not appear to be a way to ignore a specific vulnerability, so I think this is the best we can do. Thanks for making this change!

PR #88, opened 2024-03-29
RNN selector support

RyanCommits:
To support React Native Navigation screen selectors, we'll need to assign the screen name value from `layoutInfo.name` to fsAttributes, where we'll read it from the SDK.

Paired with [PR](https://github.com/cowpaths/mn/pull/74866)

### DIFF
--- a/ios/FullStory.mm
+++ b/ios/FullStory.mm
@@ -384,7 +384,7 @@ static bool array_contains_string(const char **array, const char *string) {
         if ([self respondsToSelector:@selector(layoutInfo)]) {
             id layoutInfo = [self performSelector:@selector(layoutInfo)];
             if ([layoutInfo respondsToSelector:@selector(name)]) {
-                set_fsAttribute(@{@"screen-name": [[self performSelector:NSSelectorFromString(@"layoutInfo")] name]}, (RCTView *)viewController.view);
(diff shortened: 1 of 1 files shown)


ios/FullStory.mm
```diff
@@ -372,6 +372,26 @@ + (void) load {
...
+#pragma clang diagnostic ignored "-Wundeclared-selector"
+    SWIZZLE_BEGIN_INSTANCE(RNNComponentViewController, @selector(renderReactViewIfNeeded), void) {
+        SWIZZLED_METHOD();
+        
+        UIViewController *viewController = self;
+        
+        if ([self respondsToSelector:@selector(layoutInfo)]) {
+            id layoutInfo = [self performSelector:@selector(layoutInfo)];
+            if ([layoutInfo respondsToSelector:@selector(name)]) {
+                set_fsAttribute(@{@"screen-name": [[self performSelector:NSSelectorFromString(@"layoutInfo")] name]}, (RCTView *)viewController.view);
```
martin-fs, line 387: You already retrieved `layoutInfo` from `self`:
```suggestion
                set_fsAttribute(@{@"screen-name": [layoutInfo name]}, (RCTView *)viewController.view);
```

martin-fs (approved):
One small improvement, otherwise, LGTM. Because the methods aren't init or copy-type methods, the use of respondsToSelector will not result in a leak.

PR #92, opened 2024-05-09
update the RN plugin to receive clicks from JS

JoshMiers-FS:
This PR adds a new exported module named `PrivateInterface`. On iOS, no methods are exported. On Android, `onFSPressForward` is exported.

If this method is called, then we use reflection to find a new (in the upcoming `1.49.0` FS release) `FS.reactNative_exec_onFsPressForward_direct` method and invoke the press forwarding.

Note: This also changes our minimum supported RN version to `0.66.0`.

### DIFF
--- a/android/src/legacy/java/com/FullStoryPrivateModule.java
+++ b/android/src/legacy/java/com/FullStoryPrivateModule.java
@@ -1,0 +1,23 @@
+package com.fullstory.reactnative;
+
+import com.facebook.react.bridge.ReactApplicationContext;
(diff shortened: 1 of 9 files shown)


RyanCommits (changes requested):
We should update the version dependency of `"@fullstory/babel-plugin-react-native"` in `package.json` once we merge the sister PR

ReeceLaF (approved):
LGTM!

RyanCommits (approved):

PR #93, opened 2024-05-13
Expo: Fix android server config for newer SDK

RyanCommits:
Use the Android config `serverUrl` for versions `1.37.0` and greater.


JoshMiers-FS (approved):
LGTM

PR #100, opened 2024-05-30
support custom configs

RyanCommits:
support adding custom configs to `build.gradle` and `info.plist`

### DIFF
--- a/plugin/src/__tests__/withFullStoryAndroid.test.ts
+++ b/plugin/src/__tests__/withFullStoryAndroid.test.ts
@@ -56,4 +56,19 @@ describe('Config Plugin Android Tests', function () {
     } as FullStoryAndroidProps);
     expect(result).toContain('recordOnStart false');
   });
+
+  it('Supports custom configs', async function () {
+    let result = appBuildGradle;
+    result = addFullStoryGradlePlugin(result, {
+      ...pluginConfigs,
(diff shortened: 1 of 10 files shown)


plugin/src/__tests__/withFullStoryAndroid.test.ts
```diff
@@ -56,4 +56,19 @@ describe('Config Plugin Android Tests', function () {
     } as FullStoryAndroidProps);
     expect(result).toContain('recordOnStart false');
   });
+
+  it('Supports custom configs', async function () {
+    let result = appBuildGradle;
+    result = addFullStoryGradlePlugin(result, {
+      ...pluginConfigs,
+      additionalConfigs: {
+        customBool: true,
```
JoshMiers-FS, line 65: Do we have any actual real custom configs to put in here to test with for iOS?

RyanCommits: For iOS, we use Expo's API, `withInfoPlist`, directly to edit the `info.plist`. I don't see it as necessary to test, since we're just passing in the config directly to their API, and that's managed by Expo. 

With Android gradle files, and iOS podfile, we are using regex to do string manipulation, so we have tests for them.

JoshMiers-FS (approved):
LGTM

PR #101, opened 2024-05-31
separate custom configs by platform

RyanCommits:
It makes sense to separate custom configs by platform, so that they're not being applied to both platforms at once.


JoshMiers-FS (approved):
LGTM

PR #104, opened 2024-07-10
run click handler on UI thread

RyanCommits:
Currently, calling `uiManager.resolveView(reactTag);` causes React Native to throw a `SoftException`, because it [expects](https://github.com/facebook/react-native/blob/0fd4a9447e432b8c537825f571409f56ae63dc90/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/fabric/FabricUIManager.java#L927) `FabricUIManager.resolveView` to be run on the UI thread. 

To avoid this error, we run our click handler on the UI thread.

Still needs to be tested on all the RN versions, but looking for early feedback on this.

[RN 74 Fabric Session](https://app.staging.fullstory.com/ui/KWH/session/3280449202851595977:4543370893277384711)

### Edit 7/12
[RN 66 Session ](https://app.staging.fullstory.com/ui/KWH/session/3378027069031434083:8327334863529245776)

### DIFF
--- a/android/src/main/java/com/fullstory/reactnative/FullStoryPrivateModuleImpl.java
+++ b/android/src/main/java/com/fullstory/reactnative/FullStoryPrivateModuleImpl.java
@@ -2,6 +2,7 @@ package com.fullstory.reactnative;
 
 import android.util.Log;
 import android.view.View;
+import com.facebook.react.bridge.UiThreadUtil;
 
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.UIManager;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/fullstory/reactnative/FullStoryPrivateModuleImpl.java
```diff
@@ -40,32 +41,37 @@ public static void onFSPressForward(final ReactApplicationContext context, final
...
-            view = uiManager.resolveView(reactTag);
-        } catch (Throwable t) {
-            // Silently ignore.
+        int reactTag = (int) tag;
+        if (reactTag == -1) {
             return;
         }
-        
-        if (view == null) {
+        UIManager uiManager = UIManagerHelper.getUIManager(context, ViewUtil.getUIManagerType(reactTag));
```
JoshMiers-FS, line 48: this should be wrapped in a try/catch as well, just to be safe

android/src/main/java/com/fullstory/reactnative/FullStoryPrivateModuleImpl.java
```diff
@@ -2,6 +2,7 @@
 
 import android.util.Log;
 import android.view.View;
+import com.facebook.react.bridge.UiThreadUtil;
```
JoshMiers-FS, line 5: just make sure this exists and compiles in our minimum supported RN version

RyanCommits: Confirmed compiles on RN 66. Did some testing, AFAIK this code only runs on RN 74 on Fabric

JoshMiers-FS (approved):
LGTM

PR #106, opened 2024-09-05
detect whether Fullstory is disabled before logging API errors

RyanCommits:
We wouldn't want to log API errors if Fullstory is disabled.

### DIFF
--- a/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
+++ b/android/src/main/java/com/fullstory/reactnative/FullStoryModuleImpl.java
@@ -8,10 +8,7 @@ import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import com.fullstory.FS;
 import com.fullstory.FSOnReadyListener;
-import com.fullstory.FSReason;
-import com.fullstory.FSStatusListener;
 import com.fullstory.FSSessionData;
-import com.fullstory.DefaultFSStatusListener;
 
 import java.util.UUID;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


JoshMiers-FS:
I don't think this PR tackles the actual issue the customer was seeing, which is what happens when FS is not actually added to the apk due. This logic will only be triggered when FS is part of the APK and either the server response disables FS or some other constraint is hit, which is good, but not what will fix the customer's actual issue.

RyanCommits:
> I don't think this PR tackles the actual issue the customer was seeing, which is what happens when FS is not actually added to the apk due. This logic will only be triggered when FS is part of the APK and either the server response disables FS or some other constraint is hit, which is good, but not what will fix the customer's actual issue.

The customer situation involves FS being disabled due to missing `enabledVariants` configuration. I tested this situation and this API does get triggered.

RyanCommits:
> @JoshMiers-FS The callback is triggered because [registerStatusListener](https://github.com/cowpaths/mn/blob/master/projects/fullstory/nativemobile/android/jdk/api/src/main/java/com/fullstory/FS.java#L35-L37) will call onFSDisabled if it isn't replaced by the actual implementation when FS is applied. This is the `onFSDisabled` callback rather than the `onSessionDisabled` callback.
> 
> @RyanCommits But we can't rely on this API as it is newer than the Pages API.

Any ideas on what to use to detect FS being disabled?

ReeceLaF:
> Any ideas on what to use to detect FS being disabled?

You could try seeing if there is a fullstory config file in the assets. I haven't tested this out, but something like the following would probably work if you have a way to access the application Context:
```java
AssetManager am = context.getAssets();
try {
    // Check if fullstory has been applied by checking if the fs-config.properties file exists or, for older clients, the fullstory.properties file
    if (am.list("fs-config.properties").length == 0 && am.list("fullstory.properties").length == 0) {
        isDisabled = true;
    }
} catch (Throwable t) {
    Log.e(TAG, "Could not determine if Fullstory is enabled. APIs may not function correctly.");
}
```
But you might not have Context access, and filesystem operations can be kind of slow, so another option could be:
```java
try {
    Class.forName("com.fullstory.instrumentation.Bootstrap");
} catch (ClassNotFoundException ignored) {
    isDisabled = true;
}
```
WDYT, @JoshMiers-FS

RyanCommits:
Testing Option 1, but this [context](https://github.com/fullstorydev/fullstory-react-native/blob/efe843fbf0eda3e59e238a6710d6b81dbe16a0da/android/src/turbo/java/com/FullStoryModule.java#L11) doesn't have the right information, and I'm not sure how to get to the main application context. 

Option 2 works, I'll defer to @JoshMiers-FS on which option sounds best. 

With my current solution, customers with versions `1.41` - `1.47` will still get this logged incorrectly.

JoshMiers-FS:
> > Any ideas on what to use to detect FS being disabled?
> 
> You could try seeing if there is a fullstory config file in the assets. I haven't tested this out, but something like the following would probably work if you have a way to access the application Context:
> 
> ```java
> AssetManager am = context.getAssets();
> try {
>     // Check if fullstory has been applied by checking if the fs-config.properties file exists or, for older clients, the fullstory.properties file
>     if (am.list("fs-config.properties").length == 0 && am.list("fullstory.properties").length == 0) {
>         isDisabled = true;
>     }
> } catch (Throwable t) {
>     Log.e(TAG, "Could not determine if Fullstory is enabled. APIs may not function correctly.");
> }
> ```
> 
> But you might not have Context access, and filesystem operations can be kind of slow, so another option could be:
> 
> ```java
> try {
>     Class.forName("com.fullstory.instrumentation.Bootstrap");
> } catch (ClassNotFoundException ignored) {
>     isDisabled = true;
> }
> ```
> 
> WDYT, @JoshMiers-FS

Yeah, that should work.

ReeceLaF:
> Yeah, that should work since that class has been around since the beginning and would only exist when FS was added.

And importantly has always been kept (i.e.- not obfuscated through Proguard/minification).

> Testing Option 1, but this [context](https://github.com/fullstorydev/fullstory-react-native/blob/efe843fbf0eda3e59e238a6710d6b81dbe16a0da/android/src/turbo/java/com/FullStoryModule.java#L11) doesn't have the right information, and I'm not sure how to get to the main application context.

You probably just need to use `context.getApplicationContext()`, but let's just use option 2 since we are ok with it, and it should be faster anyway.

ReeceLaF (approved):
LGTM

PR #107, opened 2024-09-10
MOCA-8077: capture uncaught exceptions

RyanCommits:
[MOCA-8077](https://fullstory.atlassian.net/browse/MOCA-8077)
Implement React Native unhandled exception logger in `eventHandler.ts`, sending stack trace information to `FS.log`. Defaulting this on `ON` to avoid manual user implementation.

[Sample session](https://app.staging.fullstory.com/ui/KWH/session/5023681300332544:1216163673456140521/devtools/console).

Open questions:
We don't have `FS.log` scrubbing in mobile. I know we scrub logs in [web](https://github.com/cowpaths/mn/blob/c511a64aa481652362bf6fae80fccf78bfe8d02b/projects/fullstory/packages/recording/src/consolewatcher.ts#L57). Is this a concern? @jurassix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,8 @@
       "license": "MIT",
       "dependencies": {
         "@fullstory/babel-plugin-annotate-react": "^2.2.0",
-        "@fullstory/babel-plugin-react-native": "^1.2.0"
+        "@fullstory/babel-plugin-react-native": "^1.3.0",
+        "@fullstory/safe-stringify": "^1.0.0"
       },
       "devDependencies": {
         "@babel/core": "^7.22.15",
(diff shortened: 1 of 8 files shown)


jurassix:
@RyanCommits the concern for ingesting logs is mostly that the data is unstructured and _could be anything_. On web we are defensive and scrub the data as a *string* looking for either Email Addresses and URLs. I think both of these are concerns for Mobile as well; especially since we now support URL scrubbing rules, those same rules can apply to the data logged or _stack frames_ if Mobile extracts those as well.

RyanCommits:
Closing until we implement log filtering on the native side.

PR #108, opened 2024-10-11
Avoid a potential initialization loop

PFedak:
VAL-8909

View controllers manage a `view`, including loading it if necessary via `loadView` when the `view` property is accessed. React Native Navigation's `RNNComponentViewController` calls `renderReactViewIfNeeded` to load the view, which typically loads both the view controller's `view` and a `reactView`.

However, it's possible for this RNN view controller to end up with a `reactView` but no `view` - my best guess is we're seeing it after UIKit has partially unloaded the view controller, but it of course doesn't know about the `reactView` field. Triggering `loadView` in such a state actually does nothing, because [there's an initial check for the presence of `reactView` ](https://github.com/wix/react-native-navigation/blob/efee2bdb8af6b0c3696c1c611f4b1492f60abbca/lib/ios/RNNComponentViewController.m#L85). Then our swizzle would end up accessing an unloaded `view`, causing an infinite loop.

We can access the view we want to give a screen name via `[UIViewController viewIfLoaded]`, avoiding possibly triggering any loader. This means we'll fail to tag the view, which isn't a problem in the situations I've been able to reproduce as the screen is about to go away, anyway, but in case there are other ways this can happen, we check `reactView`'s parent (which would normally be the `view`), too.

### DIFF
--- a/ios/FullStory.mm
+++ b/ios/FullStory.mm
@@ -252,6 +252,9 @@ static void set_dataSourceFile(id json, RCTView *view) {
 }
 
 static void set_fsAttribute(id json, RCTView *view) {
+    if (!view)
+        return;
+
     NSDictionary *newAttrs = (NSDictionary *)json;
 
     /* Clear up all the old attributes first, if they exist. */
@@ -393,10 +396,18 @@ static bool array_contains_string(const char **array, const char *string) {
         SWIZZLED_METHOD();
         
         UIViewController *viewController = self;
-        RCTView *view = (RCTView *)[viewController viewIfLoaded];
-
         if ([self respondsToSelector:@selector(layoutInfo)]) {
             id layoutInfo = [self performSelector:@selector(layoutInfo)];
+            // accessing the view controller's view directly will load the view if it isn't already,
+            // which can cause an infinite loop as RNNComponentViewController's loadView calls the
+            // function we're swizzling here
+            RCTView *view = (RCTView *)[viewController viewIfLoaded];
+            if (!view && [self respondsToSelector:@selector(reactView)]) {
+                // the view controller's init also creates reactView as a child of the main view
+                RCTView *reactView = [self performSelector:@selector(reactView)];
+                if (reactView)
+                    view = reactView.superview;
+            }
             if (view && [layoutInfo respondsToSelector:@selector(name)]) {
                 set_fsAttribute(@{@"screen-name": [layoutInfo name]}, view);
             }
--- a/ios/FullStory.mm
+++ b/ios/FullStory.mm
@@ -393,11 +393,12 @@ static bool array_contains_string(const char **array, const char *string) {
         SWIZZLED_METHOD();
         
         UIViewController *viewController = self;
-        
+        RCTView *view = (RCTView *)[viewController viewIfLoaded];
+
         if ([self respondsToSelector:@selector(layoutInfo)]) {
             id layoutInfo = [self performSelector:@selector(layoutInfo)];
-            if ([layoutInfo respondsToSelector:@selector(name)]) {
-                set_fsAttribute(@{@"screen-name": [layoutInfo name]}, (RCTView *)viewController.view);
+            if (view && [layoutInfo respondsToSelector:@selector(name)]) {
+                set_fsAttribute(@{@"screen-name": [layoutInfo name]}, view);
             }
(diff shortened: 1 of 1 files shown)


ios/FullStory.mm
```diff
@@ -393,11 +393,12 @@ + (void) load {
         SWIZZLED_METHOD();
         
         UIViewController *viewController = self;
-        
+        RCTView *view = (RCTView *)[viewController viewIfLoaded];
+
         if ([self respondsToSelector:@selector(layoutInfo)]) {
             id layoutInfo = [self performSelector:@selector(layoutInfo)];
-            if ([layoutInfo respondsToSelector:@selector(name)]) {
-                set_fsAttribute(@{@"screen-name": [layoutInfo name]}, (RCTView *)viewController.view);
```
martin-fs, line 400: Ah, I understand the issue now. Thanks for discussing with me.
If the view isn't loaded, accessing `view` causes `-[UIViewController loadViewIfRequired]` to be called, which then [causes](https://github.com/wix/react-native-navigation/blob/73a8564a83e3a9acc1748e9a2354e1396c180fe1/lib/ios/RNNComponentViewController.m#L70) `renderReactViewIfNeeded` to be called again. The thing is, the `view` should be there, since it's set by `renderReactViewIfNeeded`. Maybe it's not "loaded" because this whole chain was initiated by a call to [`loadViewIfNeeded`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621446-loadviewifneeded?language=objc), but only when the call completes is the view considered to be "loaded".

In these kind of implicit recursions, a global variable which we set upon entering the function and clear upon exiting can "cut the loop". On the surface, `viewIfLoaded` seems like a better way, but **we know the `view` property has been set at this point** (we can check to make sure) so we can bail out when we are called for a second time. However, `viewIfLoaded` may return `nil` while we're in the middle of the `loadViewIfNeeded` call.

If we don't like the idea of a global, alternately, can we swizzle `-[RNNComponentViewController viewDidLoad]` ([docs](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621495-viewdidload?language=objc)) and add `screen-name` there? @RyanCommits would have more context as to whether this would be OK (the method was discussed in [this thread](https://fullstory.slack.com/archives/C040XFG4T43/p1709824496175899)).

martin-fs (commented):
We should probably also check `view` to see if it's `nil` in `static void set_fsAttribute(id json, RCTView *view)`.

ios/FullStory.mm
PFedak: Are we confident `renderReactViewIfNeeded` will never initialize a child component view controller? I don't know enough about RN(N) to say if this is an obvious no. 

It sounds like you're discussing a scenario where some hypothetical implementation of `[RNNComponentViewController loadView]` that looks like
```objc
[self renderReactViewIfNeeded];
self.view = /* view creation logic */;
```
where the view will be created only after our swizzle returns. I don't see how we could address this situation soely by swizzling `renderReactViewIfNeeded`, as we simply never get to the view being set. 

Can you elaborate on
> we know the view property has been set at this point (we can check to make sure) 

I'm wary of what sort of assumptions we can make here - I wouldn't want to trust RNN to be safe with calling `renderReactViewIfNeeded` inside of itself if it's expecting to only be called on the main thread. Would you still call it once recursively with the global variable, or would we bail before running the swizzled method?

martin-fs: I will say that accessing the view may cause some "spooky side effects" - we know it sort of does already. So that is an argument against accessing `view`, but *we already know that renderReactViewIfNeeded sets `view`*.

Are you proposing a different way of attaching the screen name property, or do you want to drop it entirely?

PFedak: ah, I see what you're saying. I guess I was okay dropping it in this case assuming it's exceptional/a bug, but I'm maybe too willing too conclude that

martin-fs: After discussing, to avoid "spooky" shenanigans, let's access `view` via `self.reactView.superview`. If reactView or superview is nil, that's a signal that the view is off limits.

jwise: You could also potentially look up the view instance property by ivar to be absolutely certain that you are going to bypass the getter, and avoid any spooky action at a distance.

PR #110, opened 2024-11-14
WECA-1218: wait for empty JS call stack before calling view command

RyanCommits:
Currently there is an existing [bug](https://github.com/facebook/react-native/issues/47576) within React Native that causes a race condition in iOS where the view command is executed before their views are mounted, causing view commands to fail.

We seek to minimize this issue by wrapping view command execution with a `setTimeout` that will wait until the JS call stack is empty before executing the view command, which serves to give view mounting a chance to finish before view commands are called. 

There is a possibility of privacy leakage since it's possible that our view scan may read masked values after the view is mounted, but before the view command is called. However, doing a stress test with a scenario where we create 1000 views with each button tap did not result in accidental text value capture in the flatbuffer. ([session](https://app.staging.fullstory.com/ui/KWH/session/6401356677021696:4169456771433979494))

Under normal usage, we hope that this workaround will be robust enough to support code-first privacy rules while React Native works the solution.

* This PR also extracts the native command logic into a separate file.

### DIFF
--- a/ios/FullStory.mm
+++ b/ios/FullStory.mm
@@ -368,6 +368,7 @@ static bool array_contains_string(const char **array, const char *string) {
 
 @implementation FSReactSwizzleBootstrap
 + (void) load {
+	_FS_MOUNTED_REACT_NATIVE_COMPONENT = YES;
 	/* class_copyMethodList in RCTComponentData's lookup of NativeProps
 	 * can't see the propConfigs that we create in the superclass.  So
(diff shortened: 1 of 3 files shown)


RyanCommits:
https://help.fullstory.com/hc/en-us/articles/29741765803287-Disabling-batchRenderingUpdatesInEventLoop-in-React-Native-New-Architecture

PR #116, opened 2025-02-12
audit fix

RyanCommits:
The audit level was raised in https://github.com/fullstorydev/fullstory-react-native/pull/85 since at the time there was a vulnerability not yet solved by the community but was blocking our build process from passing. 

Let's lower the audit level so that we can resume checking for vulnerabilities. 

I also ran `npm audit fix` to fix existing vulnerabilities.


ScottLNorvell (approved):
lgtm

PR #121, opened 2025-04-09
VAL-9294: Provide noop API for web environments

RyanCommits:
[VAL-9294](https://fullstory.atlassian.net/browse/VAL-9294)

This work aims at avoiding module resolution errors when Expo's metro is bundling for a web environment. We create an `index.web.ts` and react native will automatically resolve to that file in a web environment, which will avoid importing native only modules.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
           key: v1-dependencies-{{ checksum "package.json" }}
       - run:
           name: Audit
-          command: npm audit --audit-level=low --production
+          command: npm audit --audit-level=low --omit=dev
       - run:
           name: Lint
           command: npm run lint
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           key: v1-dependencies-{{ checksum "package.json" }}
       - run:
           name: Audit
-          command: npm audit --audit-level=low --production
+          command: npm audit --audit-level=low --omit=dev
```
RyanCommits, line 24: Use updated command from npm

src/index.ts
```diff
@@ -187,7 +141,7 @@ export function applyFSPropertiesWithRef(existingRef?: ForwardedRef<unknown>) {
   };
 }
 
-const FullStoryAPI: FullStoryStatic = {
+const FullstoryAPI: FullstoryStatic = {
```
JoshMiers-FS, line 144: Will renaming this cause issues with existing apps?

JoshMiers-FS: I suggest we be safe and keep the existing FS casing for now.

jurassix: I think this is fine becuase of how we export this.
 
 This means the export name is `default` and not explicitly named.
 ```export default FullstoryAPI;````
 example import:
 ```inport MyName from 'package`;```
 
 It's a named export as:
 ```export FullstoryAPI;```
 example import:
 ```import { FullstoryAPI } from 'package';```

JoshMiers-FS (approved):
LGTM!

PR #122, opened 2025-04-10
Synchronize invoking the onReady callback

PFedak:
A customer experienced a rare crash that seems consistent with a race in accessing the resolve callback passed in from the javascript onReady API. If called at just the right time, the delegate could try to invoke the callback at the same time as the code setting the callback, leading to it being nulled out just before one of them calls it.


RyanCommits:
https://github.com/fullstorydev/fullstory-react-native/pull/123

Adding the `[resolve copy]` portion for extra protection

martin-fs (approved):
Analogue of https://github.com/fullstorydev/fullstory-react-native/pull/123

nfriedly (approved):
I haven't reproduced the issue, but this looks like a plausible fix, and probably a good idea regardless.

PR #123, opened 2025-04-11
race condition safety

RyanCommits:


ios/FullStory.mm
```diff
@@ -171,20 +171,26 @@ - (void) getCurrentSessionURL:(RCTPromiseResolveBlock)resolve reject:(RCTPromise
...
+        onReadyPromise(dict);
+
+        onReadyPromise = nil;
+    }
 }
 
 - (void) onReady:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject {
-  	onReadyPromise = resolve;
+    @synchronized (self) {
+        onReadyPromise = [resolve copy];
```
martin-fs, line 192: I believe [this](https://github.com/facebook/react-native/blob/0929697a6df9ab5a3654c0077bb95427eae77b4c/packages/react-native/React/Base/RCTBridgeModule.h#L38-L42) is where RCTPromiseResolveBlock is defined:

```c++
/**
 * Block that bridge modules use to resolve the JS promise waiting for a result.
 * Nil results are supported and are converted to JS's undefined value.
 */
typedef void (^RCTPromiseResolveBlock)(id result);
```

Yes, I agree that copying the block is the right thing to do here. [Blocks are Objective C objects](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html). It won't [introduce a retain cycle](https://stackoverflow.com/questions/4352561/retain-cycle-on-self-with-blocks), and our use of the block may very well outlive the scope in which the block was declared.

[Here's the Apple developer reference I found (from 2011)](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Blocks/Articles/bxUsing.html#//apple_ref/doc/uid/TP40007502-CH5-SW2) that discusses copying blocks:
> Typically, you shouldn’t need to copy (or retain) a block. You only need to make a copy when you expect the block to be used after destruction of the scope within which it was declared. Copying moves a block to the heap.
>
> You can copy and release blocks using C functions:
> ```
> Block_copy();
> Block_release();
> ```
> To avoid a memory leak, you must always balance a `Block_copy()` with `Block_release()`.

This documentation pre-dates automated reference counting for Objective C.

It turns out that, unlike essentially all other Objective C types, [blocks are allocated on the stack](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Blocks/Articles/bxVariables.html#//apple_ref/doc/uid/TP40007502-CH6-SW6), so they need to be copied to the heap if they are not going to be used immediately.

In addition, the above document says that sending the copy message just performs `Block_copy`. This makes sense, because the copy message should invoke the underlying copy functionality. So, what you're doing makes more sense than calling `Block_copy` directly.

PR #127, opened 2025-06-17
WECA-1403: fix metro import error

RyanCommits:
https://fullstory.slack.com/archives/C1DAWDPM3/p1749760541114209

**Context**
Customer reported a metro import error for version RN 72. This doesn't error doesn't exist in newer versions of React Native (79, 80), so the issue should be resolved by metro. This refactor is for backwards compatibility.

**Issue**
In `index.ts`, the existence of this import: `getInternalInstanceHandleFromPublicInstance =
    require('react-native/Libraries/ReactNative/ReactFabricPublicInstance/ReactFabricPublicInstance').getInternalInstanceHandleFromPublicInstance;` causes older metro to fail to resolve importing `LogLevel`. 
    
**Solution**
This issue seems to resolve itself if `LogLevel` is moved to a file that exports another value. This PR simply moves `global` and `isTurboModuleEnabled` to another file, so that the `LogLevel` doesn't break metro. 

Tested with React Native 66, 72, and 80.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2645,9 +2645,9 @@
       }
     },
     "node_modules/@expo/devcert/node_modules/brace-expansion": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.0.1.tgz",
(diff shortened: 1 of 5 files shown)


JoshMiers-FS (approved):
LGTM

PR #135, opened 2025-10-07
WECA-1480: auto swizzle commands

RyanCommits:


ios/FullStory.mm
```diff
@@ -338,64 +345,117 @@ - (void) set_fsAttribute:(id)json forView:(RCTView*)view withDefaultView:(RCTVie
...
-    SWIZZLE_HANDLE_COMMAND(RNCWebView);
+    SEL handleCommandSel = @selector(handleCommand:args:);
+    Class viewComponentView = NSClassFromString(@"RCTViewComponentView");
+    
+    if (!viewComponentView) {
+        NSLog(@"FullStory: RCTViewComponentView not found, cannot swizzle handleCommand");
+    } else {
+        int classCount = objc_getClassList(NULL, 0);
+        #pragma clang diagnostic push
+        #pragma clang diagnostic ignored "-Wvla-extension"
```
martin-fs, line 436: Do you know if we need to suppress this warning?

RyanCommits: Not familiar with this, but here's what AI says:

### Why We Need to Suppress It
The warning is about Variable Length Arrays (VLA), which is a C99 feature that allows you to declare arrays with runtime-determined sizes:
1. VLA is a C99 extension: While it's standard C, some compilers or strict settings flag it as an extension
2. Alternative would be more complex: Without VLA, you'd need to use malloc/free or a fixed-size array
3. This is safe usage: We get the exact count first, then allocate the array with that count
4. Common pattern: This is a standard way to get all classes from the Objective-C runtime

So seems like this is an alternative:
```
int classCount = objc_getClassList(NULL, 0);
Class *classes = malloc(classCount * sizeof(Class));
objc_getClassList(classes, classCount);
free(classes);
```

ios/FullStory.mm
```diff
@@ -338,65 +345,143 @@ - (void) set_fsAttribute:(id)json forView:(RCTView*)view withDefaultView:(RCTVie
...
+                Method handleMethod = class_getInstanceMethod(cls, handleCommandSel);
+                
+                if (handleMethod) {
+                    // Check if already swizzled using class pointer as unique key
+                    NSString *key = [NSString stringWithFormat:@"fs_swizzled_%p", cls];
+                    const char *keyStr = [key UTF8String];
+                    
+                    if (!objc_getAssociatedObject(cls, keyStr)) {
+                        swizzle_handleCommand_for_class(cls);
+                        objc_setAssociatedObject(cls, keyStr, @YES, OBJC_ASSOCIATION_RETAIN);
```
martin-fs: Don't we mark as swizzled in `swizzle_handleCommand_for_class`?

ios/FullStory.mm
```diff
@@ -338,65 +345,143 @@ - (void) set_fsAttribute:(id)json forView:(RCTView*)view withDefaultView:(RCTVie
...
+    dispatch_once(&onceToken, ^{
+        Class viewComponentView = NSClassFromString(@"RCTViewComponentView");
+        if (!viewComponentView) {
+            NSLog(@"FullStory: RCTViewComponentView not found, cannot set up lazy swizzling");
+            return;
+        }
+        
+        // Check if RCTViewComponentView has its own +initialize method
+        Method initializeMethod = class_getClassMethod(viewComponentView, @selector(initialize));
+        Method parentInitializeMethod = class_getClassMethod(class_getSuperclass(viewComponentView), @selector(initialize));
```
martin-fs: So the idea is that the lazy swizzle is installed after `[RCTViewComponentViewSubclass initialize]` is called. Interesting - [here's the documentation for initialize](https://developer.apple.com/documentation/objectivec/nsobject-swift.class/initialize()?language=objc). Seems like the right thing to do.

ios/FullStory.mm
```diff
@@ -338,65 +345,143 @@ - (void) set_fsAttribute:(id)json forView:(RCTView*)view withDefaultView:(RCTVie
...
+            }
+        });
+        
+        if (hasOwnInitialize) {
+            // Replace RCTViewComponentView's own +initialize
+            method_setImplementation(initializeMethod, newInitializeIMP);
+        } else {
+            // Add +initialize method specifically to RCTViewComponentView
+            class_addMethod(object_getClass(viewComponentView), @selector(initialize), newInitializeIMP, "v@:");
+        }
```
martin-fs: Specifically should make sure to test on classes both that have their own `initialize` method as well as classes that don't. I expect almost nobody implements their own `initialize`, though. Maybe make one with its own initialize method if there isn't one handy...

RyanCommits:
> Thanks for finding a solution to this hairy problem. If you feel like you've tested this on a sufficient number of common packages (ideally including all the ones that we had to do workarounds for) and it works then let's go with it.
> 
> [Here's a search query](https://github.com/search?q=%2F%5C%2B%5Cs%2B%5C%28void%5C%29%5Cs*initialize%2F+language%3AObjective-C&type=code&l=Objective-C) I did to look for `initialize` methods. I think [this query](https://github.com/search?q=%2F%5C%2B%5Cs%2B%5C%28void%5C%29%5Cs*initialize%2F+AND+RCTViewComponentView+language%3AObjective-C&type=code) refines it to look for files that also contain `RCTViewComponentView` and it shows no results (of course, they could exist in private code). We may want to manually double check React Native itself to satisfy ourselves that there aren't any in there. [Here's a similar query](https://github.com/search?q=%2F%5C-%5Cs%2B%5C%28instancetype%5C%29init%2F+AND+RCTViewComponentView+language%3AObjective-C&type=code) that looks for `init` methods in files with `RCTViewComponentView` and it does find examples of that.

I don't think `init` methods are related to `initialize` methods right? If someone implemented `initialize` in their private code, their view wouldn't have worked with Fullstory anyway, since we required explicit swizzling. I think this solution is a net positive and doesn't introduce any regressions.

PR #149, opened 2026-03-27
extend iOS ref rewriting test cases

RyanCommits:
Added `it.failing` test: `custom component with FS attributes does not have phantom ref injected into its props`

The plugin currently injects a synthetic ref on any element with FS attributes, including plain function components that don't forward refs to a native host instance. For these components, sharedRefWrapper is a no-op (no native view to read fs attributes to), and the injected ref leaks into `...rest` spreads, potentially overriding legitimate child refs.

The correct behavior is to only inject on components that forward refs to native host instances. Reliably detecting this at Babel transform time is non-trivial, so this is left as a known issue. The `it.failing` test documents the intended behavior and we can choose to investigate in the future.


JoshMiers-FS (approved):
LGTM

PR #151, opened 2026-06-26
iOS new architecture: migrate away from ref rewriting

RyanCommits:
Stacked with https://github.com/fullstorydev/fullstory-babel-plugin-react-native/pull/72

The previous approach to delivering FS attributes (`fsClass, fsAttribute, fsTagName, dataElement, dataComponent, dataSourceFile`) on the New Architecture worked by intercepting React element creation and replacing every element's `ref` with a wrapper function (`applyFSPropertiesWithRef`) that dispatched a `setBatchProperties` native command when React attached the ref during the layout phase.

This PR replaces that mechanism entirely with a Fabric renderer commit-phase hook.

### Changes

Not much changed in this PR. `applyFSPropertiesToInstance` behaves similarly to `applyFSPropertiesWithRef` in that it calls our native command to pass FS properties. It no longer needs to act as a wrapper to a `ref` object.

Since we are no longer wrapping `refs` we can remove some of the tests involves `refs`

---

> [!NOTE]
> **Medium Risk**
> Changes how FS attributes reach native views on iOS Fabric and depends on the paired babel plugin commit hook; misalignment could break capture or ref behavior, though old architecture still uses the view-manager path.
> 
> **Overview**
> On **iOS New Architecture**, Fullstory attribute delivery no longer rewrites element `ref`s with `applyFSPropertiesWithRef`. The public API is now **`applyFSPropertiesToInstance`**, which the babel plugin’s Fabric **commit-phase** hook calls with the committed host instance and props; it still batches `fs*` / `data*` fields into one `setBatchProperties` native command (same gating: turbo modules, iOS, non-TV).
> 
> Implementation drops ref-wrapper machinery (`FS_REF_SYMBOL`, `sharedRefWrapper`, `getInternalInstanceHandleFromPublicInstance` / Fabric public-instance reads). Tests target the new function directly and drop render/ref/unmount coverage; **`setupTests`** no longer mocks `ReactFabricPublicInstance`.
> 
> **README** and **CHANGELOG** normalize product spelling to “Fullstory” (including import example `Fullstory` from `@fullstory/react-native`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 829f4e5f5b44c88dbfa42d1088b2eb4f58a48b2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -125,22 +125,22 @@ export function applyFSPropertiesToInstance(
 
   const { fsClass, fsAttribute, fsTagName, dataElement, dataComponent, dataSourceFile } = props;
 
-  if (typeof fsClass === 'string') {
+  if (fsClass && typeof fsClass === 'string') {
     batchedProps.fsClass = fsClass;
   }
   if (fsAttribute && typeof fsAttribute === 'object') {
     batchedProps.fsAttribute = fsAttribute;
   }
-  if (typeof fsTagName === 'string') {
+  if (fsTagName && typeof fsTagName === 'string') {
     batchedProps.fsTagName = fsTagName;
   }
-  if (typeof dataElement === 'string') {
+  if (dataElement && typeof dataElement === 'string') {
     batchedProps.dataElement = dataElement;
   }
-  if (typeof dataComponent === 'string') {
+  if (dataComponent && typeof dataComponent === 'string') {
     batchedProps.dataComponent = dataComponent;
   }
-  if (typeof dataSourceFile === 'string') {
+  if (dataSourceFile && typeof dataSourceFile === 'string') {
     batchedProps.dataSourceFile = dataSourceFile;
   }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,7 @@
 import { HostComponent, NativeModules, Platform } from 'react-native';
 import codegenNativeCommands from 'react-native/Libraries/Utilities/codegenNativeCommands';
 import type { ViewProps } from 'react-native/Libraries/Components/View/ViewPropTypes';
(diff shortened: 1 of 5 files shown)


martin-fs (approved):
LGTM other than potentially keeping some sanity checks.

src/index.ts
```diff
@@ -87,131 +87,67 @@ declare type FullStoryPrivateStatic = {
...
-  if (existingRef && existingRef[FS_REF_SYMBOL]) {
-    return existingRef;
+  if (fsAttribute && typeof fsAttribute === 'object') {
+    batchedProps.fsAttribute = fsAttribute;
   }
-
-  // Use shared wrapper for null/undefined refs or static attributes
-  if (!existingRef && !hasDynamicAttributes) {
-    return sharedRefWrapper;
+  if (typeof fsTagName === 'string') {
```
martin-fs: Sorry, somehow Github ate my comment here.

Does it hurt to keep the check that the properties aren't "falsy"? The type of props is Record<string, unknown>, right? I think this is the same as before - doesn't that mean it could be undefined?

JoshMiers-FS (approved):
LGTM

src/index.ts
RyanCommits: The difference between `fsTagName && typeof fsTagName === 'string'` and `typeof fsTagName === 'string'` is that if `fsTagName` is an empty string, `typeof fsTagName` is `'string'` and would pass the check. I'll add the falsy check back to prevent us from accepting empty strings.

